PR #22, opened 2021-02-12
Feature/mc 9335 default data types

pjmonks:
Fix API call to create data model

* Pass correct value for `defaultDataTypeProvider` in `POST` request
* General refactor to include more type definitions

Note: also depends on `mdm-resources` PR https://github.com/MauroDataMapper/mdm-resources/pull/1


jamesrwelch (approved):
Happy with the iteration on this one, and trust that it's now fixed with the change to the BE

PR #62, opened 2021-04-26
mdm-select would continue to override value back to default on any ch…

gammonpeter:
…ange, removed on change to stop this happening


pjmonks (dismissed):
All good. Only thing left to fix is the merge conflict.

pjmonks (approved):

PR #64, opened 2021-04-26
Updated to remove the none required fields

gammonpeter:


src/app/wizards/dataModel/data-model-main/data-model-main.component.ts
```diff
@@ -34,11 +36,18 @@ export class DataModelMainComponent implements OnInit {
...
-  parentFolderId: any;
+  parentFolderId: string;
   parentFolder: any;
-  model: any = {
-    metadata: [],
-    classifiers: []
+  model : {
+    metadata: Array<any>;
+    classifiers: Array<any>;
+    label:string | undefined;
```
pjmonks, line 44: 👍 for adding additional types. It is also possible to say that `string | undefined` is the same as `label?: string` - they are equivalent.

I'm also trying to add a lot of these types into `mdm-resources` so that the UI does no have to keep redefining them.

src/app/wizards/dataModel/data-model-main/data-model-main.component.ts
```diff
@@ -50,6 +59,7 @@ export class DataModelMainComponent implements OnInit {
 
   ngOnInit() {
     this.title.setTitle('New Data Model');
+    this.model = { metadata :[], classifiers : [], label:undefined, description:undefined, author:undefined, organisation: undefined, dataModelType:undefined,dialect:undefined, selectedDataTypeProvider:undefined};
```
pjmonks, line 62: If something can optionally be `undefined` I don't think you have to explicitly set this on object creation. Makes no difference though.

src/app/wizards/dataModel/data-model-main/data-model-main.component.ts
```diff
@@ -105,19 +115,20 @@ export class DataModelMainComponent implements OnInit {
...
         defaultDataTypeProvider: this.model.selectedDataTypeProvider.name
       };
     }
 
-    try {
-      const response = await this.resources.dataModel.addToFolder(this.parentFolderId, resource, queryStringParams).toPromise();
+    this.resources.dataModel.addToFolder(this.parentFolderId, resource, queryStringParams).pipe(catchError(error => {
+      this.messageHandler.showError('There was a problem saving the Data Model.', error);
+      return EMPTY;
+    })).subscribe((response) => {
```
pjmonks, line 128: 👍 for using observables all the way

pjmonks (approved):
All good

PR #65, opened 2021-04-27
Modified Merge component to take a domain type prop which adjusts cal…

gammonpeter:
…ls to the BE

### DIFF
--- a/src/app/model/model-domain-type.ts
+++ b/src/app/model/model-domain-type.ts
@@ -1,3 +1,21 @@
+/*
+Copyright 2021 University of Oxford
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


src/app/model/model-merging-model.ts
```diff
@@ -0,0 +1,32 @@
...
+*/
+
+import { ModelDomainType } from './model-domain-type';
+
+export class ModelMergingModel {
+  sourceId: string;
+  targetId: string;
+  catalogueDomainType: ModelDomainType;
+
+  constructor(sourceId, targetId,catalogueDomainType)
```
pjmonks, line 26: A `class` this simple could be simplified further. Either it can just become an `interface` - meaning no constructor is required at all, so long as creating an object literal is OK for this use case - or the class can be simplified to this:

```ts
export class ModelMergingModel {
  constructor(public sourceId: string, public targetId: string, public catalogueDomainType: ModelDomainType) { }
}
```

src/app/model/model-domain-type.ts
```diff
@@ -0,0 +1,10 @@
+export enum ModelDomainType {
```
pjmonks, line 1: Missing licence header

pjmonks (changes requested):
Just a few comments I'd like you to review first please.

pjmonks (approved):

PR #76, opened 2021-04-29
Changed redirect and updated title for ref importer

gammonpeter:


src/app/import-models/import-models.component.ts
```diff
@@ -69,40 +70,54 @@ export class ImportModelsComponent implements OnInit {
     private stateHandler: StateHandlerService,
     private broadcastSvc: BroadcastService,
     private stateService: StateService
-  ) { }
+  ) {}
 
   ngOnInit() {
     // tslint:disable-next-line: deprecation
-    this.importType = this.stateService.params.importType ? this.stateService.params.importType : 'dataModels';
+    this.importType = this.stateService.params.importType
```
pjmonks, line 77: While modifying this area, could you also replace the deprecated `stateService.params` with `uiRouterGlobal.params` please?

pjmonks (changes requested):
The actual fix looks good. I left one comment to tidy up the changes a bit more, if you could do that please then I'll merge for you.

src/app/import-models/import-models.component.ts
gammonpeter: Updated to use UiRouterGlobals

pjmonks (approved):
All good

PR #78, opened 2021-05-04
Added Styling and corrected issue where ui wouldn't update

gammonpeter:


src/app/referenceData/reference-data-details/reference-data-details.component.html
```diff
@@ -253,6 +253,7 @@
               [hideDelete]="true"
               [hideCancel]="false"
             ></mdm-editable-form-buttons>
+            <sup *ngIf="result?.readableByEveryone" class="text-muted" style="margin-left: 4px;"><i class="fas fa-globe-europe globe fa-xs" aria-hidden="true" matTooltip="Publicly Available"></i></sup>
```
pjmonks, line 256: When I was working on custom branding, I found that there were CSS classes already defined for adding margins/padding to introduce spacing. The selectors followed a pattern like:

* `ml-1` represents `margin-left: 4px`
* `mt-2` represents `margin-top: 8px`
* `pr-3` represents `padding-right: 12px`
* Number increments went up by `4px`
* And so on

The full list of selectors can be found in `src\style\abstracts\_helpers.scss`. You don't need to change this for now, but maybe keep in mind for the future.

src/app/terminology/terminology-details/terminology-details.component.ts
```diff
@@ -132,7 +154,7 @@ export class TerminologyDetailsComponent implements OnInit {
     this.title.setTitle(`Terminology - ${this.mcTerminology?.label}`);
   }
 
-  validateLabel = data => {
+  validateLabel = (data) => {
```
pjmonks, line 157: Andrew Tsui suggested that functions written in this property syntax should be changed back to their standard function syntax for consistency, e.g. `validateLabel(data) { }`.

Also, if you happen to know the type of `data` then I'd suggest adding the definition to the parameter as well.

src/app/terminology/terminology-details/terminology-details.component.ts
```diff
@@ -283,78 +340,114 @@ export class TerminologyDetailsComponent implements OnInit {
...
+              '<strong>Note: </strong>All its \'Terms\' will be deleted <span class=\'warning\'>permanently</span>.'
+          }
         }
-      })
+      )
       .subscribe(() => this.delete(true));
   };
 
-  openEditClicked = formName => {
+  openEditClicked = (formName) => {
```
pjmonks, line 366: See previous comment about property vs function syntax

pjmonks (changes requested):
Implementation is perfectly fine. I just left a couple of feedback items to tidy up the code a bit.

pjmonks (approved):
👍

PR #84, opened 2021-05-05
Feature/mc 9307 folder screens

pjmonks:
Updated the folder views to match the new UI layout, plus allow viewing/editing of the following:

* Profiles
* Annotations
* Attachments
* History
* Rules


gammonpeter (approved):

gammonpeter:
Issue with the profiles i believe is to do with the api issues approving

PR #85, opened 2021-05-07
Added 504 error status page and specific call to it

gammonpeter:


src/app/app-routing.module.ts
```diff
@@ -122,6 +123,11 @@ export const pageRoutes: { states: Ng2StateDeclaration[] } = {
       url: '/serverError',
       component: ServerErrorComponent
     },
+    {
+      name: 'appContainer.mainApp.twoSidePanel.catalogue.serverTimeout',
+      url: '/serverError',
```
pjmonks, line 128: Should we use a different url for the new server timeout page?

src/app/errors/server-timeout/server-timeout.component.ts
```diff
@@ -0,0 +1,46 @@
...
+export class ServerTimeoutComponent extends ErrorComponent implements OnInit {
+
+    constructor(protected messageService: MessageService,
+                protected clipboardService: ClipboardService,
+                protected sharedService: SharedService,
+                protected youtrackService: YoutrackService) {
+        super(messageService, clipboardService, sharedService, youtrackService);
+        this.errorHeader = 'Server Timeout Error';
+        this.errorMessage = 'We\'re sorry, but the server responded with an error message.';
+        this.errorResolution = 'The request may of succeeded, please allow sometime and then select the catalogue item you were on from the tree.';
```
pjmonks, line 38: Can I suggest slightly rewording this (emphasis added for changes)?

> The request may _have_ succeeded. _Please_ allow _some time_ and then select the catalogue item you were on from the _model_ tree _to refresh the view_.

gammonpeter: Updated wording

PR #95, opened 2021-05-11
Updated Model Merging to correct issue with dev

gammonpeter:


src/app/model-merging/model-merging.component.ts
```diff
@@ -506,14 +502,14 @@ export class ModelMergingComponent implements OnInit {
     if (!labelDiff.isMergeConflict) {
       update['acceptSource'] = true;
       update['acceptTarget'] = false;
-      this.onAcceptPress(update, 'source', 'property');
+      this.onAcceptPress(update, this.DIFF_LOC.Source, 'property');
```
pjmonks, line 505: I like the use of a constant here.

You don't need to change anything, but another alternative I discovered is to use discriminated unions e.g.

```ts
function onAcceptPress(diff: any, diffLocation: 'source' | 'target', type: string) { }
```

This would allow a string but only allow 'source' or 'target' - anything else is a compile error. 

Don't change what you've done here though, I'm happy with it.

pjmonks (approved):
Looks good, I like the tidying up you've done.

PR #119, opened 2021-05-17
Added Annotations tab to all requied windows

gammonpeter:


pjmonks (changes requested):
All looks good, except that folders don't seem to be updated - they still have separate Attachments/Comments tabs which looks inconsistent, could you also update those too please?

Another note, but may be more related to a bigger piece of work, is that I'm noticing a lot of duplication in the parent model views for all these tabs, especially copying similar markup for flex layouts. Do you think there should be a better way of defining these common tabs? Not for this PR/issue, but for future.

pjmonks (approved):

PR #125, opened 2021-05-17
Add file license checking to jenkins

olliefreeman:
Update the excluded files list to include all the folders we don't want and to properly exclude the files using ant glob patterns
Add the check command to Jenkins with an unstable warning, this way any new files with the wrong license will cause the build to fail.


pjmonks (approved):
Can't see any issues with this myself

PR #127, opened 2021-05-18
Help topics navigate to Mauro documentation site

### LINKED ISSUE #101: Terminology Help show Not Found Error
Click on a terminology and select help it shows a not found error

pjmonks:
Fixes #101 

* Replace modal help dialog that uses portions of the old wiki with links that navigate to a new browser tab to the relevant page in Mauro Data Mapper documentation site
* Update footer to use correct link
* Manage help topics via the 'environment.ts` file now
* Remove any help icons where old help topics do not have an equivalent documentation page


gammonpeter (approved):

theresanoble:
@pjmonks having a link on the home page before you log in would be very very helpful for new users. I wasn't aware there was new documentation until Ollie sent me the link this morning and I've been using MDM actively for weeks now.... Just something a bit bolder so that as a user you see it I think would be helpful for those are aren't completely comfortable in the metadata world. (if that makes sense)

PR #138, opened 2021-05-21
Updated enum value to be correct, changed to use mdm resouces and add…

gammonpeter:
…ed missing elements to child data classes

### DIFF
--- a/src/app/services/handlers/security-handler.service.ts
+++ b/src/app/services/handlers/security-handler.service.ts
@@ -27,8 +27,7 @@ import { AuthenticatedSessionError, SignInError, UserDetails } from './security-
 import { Observable, of, throwError } from 'rxjs';
 import { HttpErrorResponse } from '@angular/common/http';
 import { catchError, map, switchMap } from 'rxjs/operators';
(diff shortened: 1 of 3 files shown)


src/app/services/handlers/security-handler.service.ts
```diff
@@ -320,22 +319,22 @@ export class SecurityHandlerService {
...
 
   elementAccess(element) {
-    if (element.domainType === DOMAIN_TYPE.DataModel ||
-      element.domainType === DOMAIN_TYPE.Terminology ||
-      element.domainType === DOMAIN_TYPE.CodeSet ||
-      element.domainType === DOMAIN_TYPE.ReferenceDataModel) {
+    if (element.domainType === CatalogueItemDomainType.DataModel ||
+      element.domainType === CatalogueItemDomainType.Terminology ||
+      element.domainType === CatalogueItemDomainType.CodeSet ||
+      element.domainType === CatalogueItemDomainType.ReferenceDataModel) {
```
pjmonks, line 325: I like this, starting to phase out types that should belong to `mdm-resources` for other clients to use.

pjmonks (approved):

PR #140, opened 2021-05-25
Feature/gh 139

gammonpeter:


src/app/code-set/code-set-details/code-set-details.component.html
```diff
@@ -154,37 +152,30 @@
...
             >
               <i matTooltip="Code Set" class="fas fa-list"></i>
             </span>
             <mdm-inline-text-edit
-              [readOnly]="showEditDescription"
-              [inEditMode]="editableForm.visible"
-              [(ngModel)]="result.label"
+              [readOnly]="false"
+              [inEditMode]="editMode"
+              [showButtons]="true"
```
pjmonks, line 166: Adding inline save/cancel buttons to the `<mdm-inline-text-edit>` is a good idea, makes the label editor in the model detail less bulky.

src/app/code-set/code-set-details/code-set-details.component.html
```diff
@@ -283,19 +274,19 @@
 
 <div
   *ngIf="
-    showSecuritySection && !editableForm.visible && !editableForm.deletePending
+    showSecuritySection && !editMode && !deleteInProgress
```
pjmonks, line 277: Security sections `<div>` elements shouldn't be needed anymore now that the user access/permissions opens a modal dialog.

src/app/code-set/code-set-details/code-set-details.component.ts
```diff
@@ -38,225 +29,135 @@ import { BroadcastService } from '@mdm/services/broadcast.service';
...
-      this.editableForm.validationError = false;
-      this.errorMessage = '';
-      this.editableForm.description = this.result.description;
-      if (this.result.classifiers) {
-        this.result.classifiers.forEach(item => {
-          this.editableForm.classifiers.push(item);
-        });
+    this.subscription = this.messageService.changeUserGroupAccess.subscribe(
+      (message: boolean) => {
+        this.showSecuritySection = message;
```
pjmonks, line 94: This might need changing, using a subscription service is probably overkill. My idea was to have the parent component load the detail/access properties and pass them down to the child component via `@Input()`, if changes need detecting then maybe use `@Output()` and/or `ngOnChanges()`.

Defer this to a new clean up issue, just writing down as a reminder.

src/app/code-set/code-set-details/code-set-details.component.ts
```diff
@@ -38,225 +29,135 @@ import { BroadcastService } from '@mdm/services/broadcast.service';
...
-            this.compareToList.push(link.target);
-          }
-        });
-      }
-
-      if (this.result != null) {
-        this.hasResult = true;
-        this.watchDataModelObject();
+  codeSetDetails(): any {
+    this.subscription = this.messageService.dataChanged$.subscribe(
```
pjmonks, line 100: Same issue with this, using `@Input()` would be much simpler if the data just comes from the parent component.

Defer to another clean up issue.

src/app/code-set/code-set-details/code-set-details.component.ts
```diff
@@ -335,115 +276,61 @@ export class CodeSetDetailsComponent implements OnInit, OnDestroy {
...
-
-    this.dialog
-      .openDoubleConfirmationAsync({
-        data: {
-          title: 'Delete permanently',
-          okBtnTitle: 'Yes, delete',
-          btnType: 'warn',
-          message: 'Are you sure you want to <span class=\'warning\'>permanently</span> delete this Code Set?'
+  save() {
+    if (this.validatorService.validateLabel(this.codeSetDetail.label)) {
```
pjmonks, line 280: I'm wondering if validation could be done somehow on the `<mdm-inline-text-edit>` component instead, would maybe look nicer to have the validation/error message inline with the control as you change the form.

pjmonks (commented):
I think this is a great start so far. The comments I've left for the moment are further clean up issues really as reminders or discussion points. Changing names to make more sense I like to, we'll work out the best way to consolidate other work as we go.

src/app/term/term-details/term-details.component.ts
```diff
@@ -16,248 +16,182 @@ limitations under the License.
...
+    if (this.validatorServive.validateLabel(this.mcTerm.definition)) {
+
+      this.resourcesService.term
+        .update(this.terminologyId, this.mcTerm.id, this.mcTerm)
+        .subscribe(
+          () => {
+            this.messageHandler.showSuccess('Term updated successfully.');
+            this.editingService.stop();
+            this.editMode = false;
+            this.errorMessage = '';
```
pjmonks, line 128: I've noticed this `errorMessage` property in these detail views, but does it actually have a purpose anymore?

src/app/terminology/terminology-details/terminology-details.component.ts
```diff
@@ -53,46 +57,26 @@ import { ModalDialogStatus } from '@mdm/constants/modal-dialog-status';
   styleUrls: ['./terminology-details.component.sass']
 })
 export class TerminologyDetailsComponent implements OnInit, OnDestroy {
-  @Input() mcTerminology: any;
-  @Input() hideEditButton: boolean;
-  @Output() afterSave = new EventEmitter<any>();
-  @Output() openEditFormChanged = new EventEmitter<any>();
-  @ViewChild('aLink', { static: false }) aLink: ElementRef;
+  @Input() mcTerminology: ModelableDetail & Modelable & Branchable & Versionable & SecurableModel;
```
pjmonks, line 60: Doesn't `mdm-resources` have a `TerminologyDetail` type to use? Not sure why this is re-combining traits that `TerminologyDetail` should already have.

src/app/terminology/terminology-details/terminology-details.component.ts
```diff
@@ -196,39 +133,20 @@ export class TerminologyDetailsComponent implements OnInit, OnDestroy {
         );
       }
     }
-  };
+  }
 
-  formBeforeSave = () => {
+  save = () => {
```
pjmonks, line 138: Might as well update this function declaration to be `save() { ... }` instead of property syntax.

pjmonks (commented):
Still looking good so far

pjmonks (commented):
Still looking good 👍

src/app/services/handlers/security-handler.service.ts
```diff
@@ -236,7 +237,8 @@ export class SecurityHandlerService {
       );
   }
 
-  dataModelAccess(element: Securable & Finalisable) {
+
+  elementAccess(element) : Access {
```
pjmonks, line 241: Is it possible to add a type to `element`? `Securable` gives you the `availableActions` property. For this bit:

```ts
showNewVersion: element.finalised
```

`element` should be `Finalisable`. So maybe try a union?

```ts
elementAccess(element: Securable | Securable & Finalisable): Access;
```

But I'm just musing out loud, don't know if that works.

pjmonks: Also, do we want to link this PR to issue #143 as well? The PR would then close both at once.

pjmonks (changes requested):
I'm impressed with the performance boost now that all this code has been stripped back.

There is one functional issue I found which needs looking at, and then some minor ones to look over.

pjmonks (approved):
I'm happy with everything and as discussed we'll leave the user/group role dialog issue for #144. Nice work! 👍

PR #147, opened 2021-05-27
Feature/gh 142 BroadcastService is strongly typed

### LINKED ISSUE #142: Refactor BroadcastService to be type safe
# Issue

The current `BroadcastService` broadcasts messages to other unrelated components as follows:

* Send a message name as a `string`
* Send optional data with the message as `any` data

This does introduce technical debt because of the copy/paste aspect of broadcasting/subscribing to these events. For example, to trigger the model tree to reload:

```ts
service.broadcast('$reloadFoldersTree');
```

If this were mispelled or the message name was changed in the future, re-testing this would have to be manual instead of found during compile time.

# Suggestion

It is possible to create a type-safe broadcast service where:

* Message names are defined as constants
* Expected data can be defined with generics

An example of interface for this could be:

```ts
export enum BroadcastEvent {
  ReloadFolderTree,
  // etc...
}

export class BroadcastMessage<T> {
  constructor(public event: BroadcastEvent, public data?: T) { }
}

export class BroadcastService {
  private handler = new Subject<BroadcastMessage<any>>();

  // Call this to subscribe to certain events
  on<T>(event: BroadcastEvent): Observable<T> { }

  // Call this to dispatch/broadcast events
  dispatch<T>(event: BroadcastEvent, data?: T): void { }
}
```

# Tasks

- [x] Refactor the `BroadcastService` to use the type-safe methods
- [x] Add the necessary `BroadcastEvent` enums. Search the codebase for all instances of the string name usage as a start
- [x] Replace the old string broadcast names with `BroadcastEvent`
- [x] Ensure all subscription points still work

pjmonks:
Resolves #142 

* `BroadcastService` is not strongly typed and has functions for dispatching and watching for common events, such as reloading the catalogue tree
* Event names are forced to be a union type to guarantee that the correct name is used
* Updated all call sites to use the new interface
* During this refactor I found some old components that were not used anymore, these have been removed too


gammonpeter (approved):
Amazing work like it

PR #151, opened 2021-05-28
Added check and redirect requesting login

gammonpeter:


pjmonks (approved):
All good, and looks simpler than what we discussed.

Issue #152, opened 2021-06-02
Public indicator not showing on all catalogue items


gammonpeter:
Public indicator not showing on all catalogue items

PR #153, opened 2021-06-02
#152 Organise readable icons in each detail view

pjmonks:
Also add "Authenticated users only" icon/status


gammonpeter (approved):
All good but maybe might be a candidate for component as it seems to repeat a lot of code

PR #158, opened 2021-06-03
Feature/gh 154 Reorder tabs

### LINKED ISSUE #154: Re-order tabs
'Annotations' Tab shouldn't be so prominent.
I'd go for the following order:

- Description
- Schema (Elements)
- Types (DM only)
- Context
- Rules
- Annotations
- History

pjmonks:
Fixes #154 

Re-order tabs so that "Annotations" is less prominent, usually one of the last ones in each tab view.

Also completed some refactoring work:

* All views have consistent tab layouts and behaviour
* Refactored tab index/name state to reduce duplicate code


gammonpeter (approved):
very impressive work

PR #162, opened 2021-06-04
mc-9480 Update the merge model diagram to handle the changes in the j…

olliefreeman:
mc-9480 Update the merge model diagram to handle the changes in the json from the API

The API wasnt using the gson templates and was therefore returning the incorrectly named fields

* `modelId` is now `id`
* `branchName` is now `branch`
* all booleans are now `isXxxx`
* only the base model is actually "not new model version of" therefore we can ignore that when determining box content

* Make sure the branch name is used for a branch and the model version is used for finalised models


olliefreeman:
This is required to be merged before we can release with the existing backend

gammonpeter (approved):

PR #164, opened 2021-06-07
MC-9515 Update model tree drag and drop

pjmonks:
* Versioned Folder support added
* Restrictions set so that Versioned Folders cannot be moved into other Versioned Folders

@gammonpeter - the code file has been formatted so finding the actual changes looks difficult. All changes were made in the `drag...()` functions inside `folders-tree.component.ts`


gammonpeter (approved):

gammonpeter (changes requested):
Lint issues but look good a works well

gammonpeter (approved):

Issue #163, opened 2021-06-07
Profile Dashboard view showing wrong namespace


jamesrwelch:
The dynamic profile dashboard view on the admin interface is showing the wrong namespaces - it's currently showing the field: `namespace`, but should in fact be showing `metadataNamespace`

PR #165, opened 2021-06-07
Update Dynamic Profile Dashboard (#163)

jamesrwelch:
Fixed the metadata field used.  Changes to the back-end will also select the correct version number


pjmonks (approved):
All good 👍

PR #169, opened 2021-06-08
MC-9513 Fix styling issues in new folder dialog

pjmonks:


gammonpeter (dismissed):
Once the jekins issue are fixed we can commit

gammonpeter (approved):

PR #176, opened 2021-06-11
Feature/mc 9518 Context menu support for Versioned Folders

pjmonks:
Also includes changes for [MC-9514 - Set root node](https://metadatacatalogue.myjetbrains.com/youtrack/issue/MC-9514)

* Re-written context menu in folder tree to use new `availableActions` field returned in `MdmTreeItem` elements. These govern (from the backend) what features should be available for each item.
* Refactored a lot of the tree component and related components
* Updated creation wizards to allow Versioned Folders to be a parent element
* Added visual updates to the context menu
* Added preliminary support for setting (versioned) folders as root nodes in the tree to focus. Note: there is a backend issue which prevents this full though, recorded in MauroDataMapper/mdm-core#84


src/app/shared/models/models.model.ts
```diff
@@ -0,0 +1,46 @@
...
+export interface MdmTreeLevelManager {
+  current: number;
+  currentFocusedElement?: MdmTreeItem;
+
+  backToTree: () => void;
+  focusTreeItem: (node: MdmTreeItem) => void;
+}
+
+export const mapCatalogueDomainTypeToContainer = (domain: CatalogueItemDomainType): ModelDomainType | undefined => {
+  if (domain === CatalogueItemDomainType.DataModel) {
```
gammonpeter, line 31: This could be cleaner as a switch statement

gammonpeter (approved):

PR #189, opened 2021-06-17
Feature/mc 9549 Enable Branch/Version/Fork for Versioned Folders

pjmonks:
* Add actions to navigate to `NewVersionComponent` from a Versioned Folder
* Tidy up the UI for the `NewVersionComponent`, especially text to make sure instructions make sense for both models and containers
* Created shared `BranchSelectorComponent` to use correct endpoint for getting version/branch list from a model/folder and update all detail views to use this now

**Note**: there is a known issue in the backend which prevents a new version/branch/fork being created for a Versioned Folder that has child content, see MauroDataMapper/mdm-core#88. Oddly, these actions do work on empty Versioned Folders though.

### DIFF
--- a/src/app/shared/new-version/new-version.component.ts
+++ b/src/app/shared/new-version/new-version.component.ts
@@ -54,7 +54,7 @@ export class NewVersionComponent implements OnInit {
       icon: 'fa-list',
       getDescription: (domainType, item) => {
         return `This will create a copy of <b>${item.label}</b> with a new name, and a new 'main' branch.
-          Use this option if you are planning on taking this ${domainType === CatalogueItemDomainType.VersionedFolder ? 'folder' : 'model'} in a new direction, or under a new authority.`
+          Use this option if you are planning on taking this ${domainType === CatalogueItemDomainType.VersionedFolder ? 'folder' : 'model'} in a new direction, or under a new authority.`;
       }
     },
     {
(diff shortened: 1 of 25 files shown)


src/app/shared/new-version/new-version.component.ts
```diff
@@ -16,185 +16,191 @@ limitations under the License.
...
-          },
-          (error) => {
-            this.processing = false;
-            this.messageHandler.showError(
-              'There was a problem creating the new Document Model version.',
-              error
-            );
-          }
-        );
+      const request = this.resources[this.domainElementType.resourceName]
```
gammonpeter, line 144: Switch Case would be cleaner here

gammonpeter (commented):
I like the use of single component for branch selector, not too sure about the changes the type selector for new version. I'd recommend James look before I merge it into Dev

src/app/shared/new-version/new-version.component.ts
pjmonks: After this comment I tried building up some adapter services around the `MdmResourcesService` which would expose common endpoint functions e.g. `newBranchModelVersion()` and then use `switch-case` statements to directly drill down to the correct resource based on domain type.

After trying this though it felt like a lot of extra code to write when the original way of property indexing was shorter and faster. Bear in mind that every endpoint with a similar signature has to have their own `switch-case` block. I'm also a bit concerned about how this would minimise during a production build, it could bloat the final distributable output.

I would suggest leaving this for now, but maybe add a new issue under the "Clean up" label/project to discuss later.

gammonpeter: agreed

pjmonks:
I've updated the dropdown selection to include the help text with each action:

image

Hopefully this is going to be a better UX for the user, to guide them to the right action.

gammonpeter (approved):

PR #194, opened 2021-06-18
Feature/mc 9541

gammonpeter:


src/app/directives/resizable.directive.ts
```diff
@@ -0,0 +1,105 @@
...
+
+
+
+    const newWidth = (wid) => {
+      const newWidthCalc = Math.max(this.resizableMinWidth, wid);
+      this.renderer.setStyle(this.el.nativeElement,'width',`${newWidthCalc}px`);
+    };
+
+
+    const mouseMoveG = (evt) => {
```
pjmonks, line 44: Does the "G" in these names refer to "Global"? If so could we rename them to `mouseMoveGlobal()` etc to make them clearer?

src/app/directives/resizable.directive.ts
```diff
@@ -0,0 +1,105 @@
...
+SPDX-License-Identifier: Apache-2.0
+*/
+
+import { Directive, ElementRef, OnInit, Input, Renderer2 } from '@angular/core';
+
+@Directive({
+  selector: '[mdmResizable]' // Attribute selector
+})
+
+export class ResizableDirective implements OnInit {
```
pjmonks, line 26: Directive code file could do with some standardised formatting e.g. remove excessive line spacing, consistent indenting etc.

src/app/merge-diff/merge-diff-container/merge-diff-container.component.html
```diff
@@ -16,4 +16,72 @@
...
 SPDX-License-Identifier: Apache-2.0
 -->
-<p>merge-diff-container works!</p>
+<div class="loading-spinner" *ngIf="!loaded">
+  <div>
+    <i class="fas fa-sync-alt fa-spin"></i>
+  </div>
+</div>
+<div  *ngIf="loaded">
+  <div class="heading-container">
```
pjmonks, line 25: Using the `heading-container` CSS class means the right-side controls, such as the "Commit" button, don't look in line with the rest of the page container:

image

Would it look better to line everything up by using the `container` class for all block elements?

src/app/merge-diff/merge-diff-container/merge-diff-container.component.scss
```diff
@@ -15,4 +15,31 @@ See the License for the specific language governing permissions and
 limitations under the License.
 
 SPDX-License-Identifier: Apache-2.0
-*/
\ No newline at end of file
+*/
+
+.container {
```
pjmonks, line 20: I thought `container` just just a standard Bootstrap layout class, why is this created again? Or does it override some CSS properties?

src/app/merge-diff/merge-diff-container/merge-diff-container.component.ts
```diff
@@ -32,9 +36,14 @@ import { SharedService, StateHandlerService } from '@mdm/services';
 })
 export class MergeDiffContainerComponent implements OnInit {
 
+  loaded = false;
+  domainType: string;
```
pjmonks, line 40: Maybe use `CatalogueItemDomainType` instead (unless there is a specific reason it's a generic string).

src/app/directives/resizable.directive.ts
pjmonks, line 26: This directive looks like a good idea.

src/app/merge-diff/merge-diff.module.ts
```diff
@@ -22,10 +22,13 @@ import { CommonModule } from '@angular/common';
...
 import { MaterialModule } from '@mdm/modules/material/material.module';
 import { MergeDiffContainerComponent } from '@mdm/merge-diff/merge-diff-container/merge-diff-container.component';
+import { ResizableDirective } from '@mdm/directives/resizable.directive';
+
 
 @NgModule({
   declarations: [
-    MergeDiffContainerComponent
+    MergeDiffContainerComponent,
+    ResizableDirective
```
pjmonks, line 31: Fine here for now, but I imagine this could be moved to a more common module such as `SharedModule`

pjmonks (changes requested):
Good start. We should probably discuss tidying this up and moving this forward together, but I could merge into `develop` if you want us to build this up together.

src/app/merge-diff/merge-diff-container/merge-diff-container.component.scss
gammonpeter: override for the resize to work

src/app/merge-diff/merge-diff-adapter/merge-diff-adapter.service.ts
```diff
@@ -27,8 +29,15 @@ import { Injectable } from '@angular/core';
...
 export class MergeDiffAdapterService {
 
-  constructor() { }
+  source: any;
 
+  constructor(private resources : MdmResourcesService) { }
+
+  loadCatalogueItem(id: string, domainType: ModelDomainRequestType)
+  {
+   return this.resources[domainType.toString()].get(domainType,id);
```
pjmonks, line 38: Perhaps `ModalDomainRequestType` has the correct names for `MdmResourcesService` properties, but the safest way I've found is to use the `ElementTypesService.getBaseTypeForDomainType()` function to return a type object that contains the correct resource property name.

src/app/merge-diff/merge-diff-container/merge-diff-container.component.ts
```diff
@@ -53,8 +57,9 @@ export class MergeDiffContainerComponent implements OnInit {
     }
 
     // const sourceId = this.uiRouterGlobals.params.sourceId;
-    // const targetId = this.uiRouterGlobals.params.targetId;
+   //  const targetId = this.uiRouterGlobals.params.targetId;
     this.domainType = this.uiRouterGlobals.params.catalogueDomainType;
+    this.source = this.mergeService.source;
```
pjmonks, line 62: Is a data model passed into the `MergeDiffContainerComponent` via the service to avoid further backend calls? If so, not sure how I feel about this method of passing the input, seems like it relies too much on knowing how to setup the component/route to the page before being able to use it.

I think `UIRouter` is able to pass objects to the component states during transitions, so that might be an alternative. If I've researched correctly then it should be something like:

```ts
// Route
{
      name: 'route',
      url: '/my-page',
      component: MyComponent,
      params: {
        source: null  // default value
      }
    }

// Transition
stateSerivce.go('route', { source: this.dataModel });   // Send object as param to state

// Access
const source = this.uiRouterGlobals.params.source;
```

pjmonks (changes requested):
Looking better

src/app/merge-diff/merge-diff-container/merge-diff-container.component.ts
gammonpeter: i'm not sure either way is the best way. It was just a way to try and cut a call out but it doesn't work on the refresh i'll remove it

pjmonks (approved):
All good

PR #197, opened 2021-06-21
Feature/gh 195 Replace build environment variables with API properties

pjmonks:
* Control feature toggles in the UI with the API properties found under the "Admin > Configuration > Properties" list.
* `environment.ts` still provides the initial defaults for these feature toggles until an API Property is added to override them
* Removed the environment variables from the build process, now obsolete


olliefreeman (approved):
Assuming this has been tested under prod build circumstances not just ng serve then im happy with it...certainly makes life easier

pjmonks:
> Assuming this has been tested under prod build circumstances not just ng serve then im happy with it...certainly makes life easier

@olliefreeman Yes, I've tested it on my local machine but from production build output served on a local server.

olliefreeman:
thats cool...im happy

PR #202, opened 2021-06-23
Moved Bootstrap CSS reference from index.html into styles.scss

atpw25:
To avoid situations where the client browser have no access to the Bootstrap CDN (e.g. offline development, or intranet environment without internet access), which causes UI layout/styling issues.


pjmonks (dismissed):
Looks good 👍

pjmonks (approved):

PR #206, opened 2021-06-28
Feature/mc 9574 Login screen supports OpenID Connect authentication

pjmonks:
* List any OpenID Connect providers under the usual username/password form, if feature is enabled.
* Redirect to external provider site for authentication
* Allow redirect back to Mauro UI and automatically send OpenID session state to usual `/api/authentication/login` endpoint to complete authentication and authorization


src/app/modals/login-modal/login-modal.component.scss
```diff
@@ -0,0 +1,37 @@
...
+.mdm-login-modal {
+
+  mat-divider {
+    margin: 4px 0;
+  }
+
+  &__providers {
+
+    &--scrollable {
+      max-height: 200px;
```
gammonpeter, line 29: i'd recommend making this proportion of the screen height this would allow smaller screens to show without issue

gammonpeter (changes requested):
good work, only make the change if you agree

src/app/modals/login-modal/login-modal.component.scss
pjmonks: Agreed, tried to make adjustments.

gammonpeter (approved):

PR #210, opened 2021-06-30
Feature/mc 9542

gammonpeter:


src/app/merge-diff/merge-diff-adapter/merge-diff-adapter.service.ts
```diff
@@ -27,8 +31,39 @@ import { MdmResourcesService } from '@mdm/modules/resources';
...
+  //      mergeDiff(sourceId, targetId,targetId)
+  //     .pipe(
+  //       catchError(() => {return EMPTY;}),  // Common error handling (maybe)
+  //       map((response: VersioningResponse) => response.body)  // Map to types we care about
+  //     );
+  // }
+
+  getMergeDiff() : Observable<any>
+  {
+    return this.http.get('../../../assets/newStyle.json');
```
pjmonks, line 64: If this is a local file there is another way to load in JSON from file, without needing this `HttpClient` injected.

Basically you can import a JSON file just like a JS/TS file/module:

```ts
import * from './test-file.json`
```

I think our Angular/TypeScript config files allow this, if not there may be some switches involved.

The JSON is then imported just like an exported object literal.

pjmonks (approved):
All looks good so far. Fix the tests/Jenkins build and we can merge.

src/app/merge-diff/merge-diff-adapter/merge-diff-adapter.service.ts
gammonpeter: this will be removed once the BE is ready.

pjmonks (approved):

PR #222, opened 2021-07-07
Feature/mc 9541

gammonpeter:


src/app/merge-diff/merge-comparsion/merge-comparison.component.html
```diff
@@ -21,78 +21,158 @@
...
+          Link Scrolls
+        </button>
+        <button
+          mat-icon-button
+          type="button"
+          matToolTip="Link Scrolls"
+          aria-label="Link Scrolls"
+          *ngIf="!isCommitting && linkScroll"
+          class="button"
+          (click)="linkScrolls(false)"
```
pjmonks, line 94: I like this feature

src/app/merge-diff/merge-comparsion/merge-comparison.component.scss
```diff
@@ -17,21 +17,36 @@ limitations under the License.
 SPDX-License-Identifier: Apache-2.0
 */
 
-.sourceCol
-{
-        height: 25vh;;
+.sourceCol {
```
pjmonks, line 20: With the component Sass files, I tend to include a top-level class and have others be children of that. For example, for this component I might have:

```sass
.mdm-merge-comparison {
    .textArea { }

    .sourceCol { }

    //etc
}
```

```html
<div class="mdm-merge-comparison">
    <div class="sourceCol">
        <!-- ... -->
    </div>
</div>
```

I do this so that there are no conflicts between other global classes e.g. there might be another `.spacer` class.

Also, I learned from from frontend developers I used to work with that [BEM naming](http://getbem.com/naming/) of CSS classes is one standard often used which may be useful.

src/app/merge-diff/merge-comparsion/merge-comparsion.component.ts
```diff
@@ -16,75 +16,143 @@ limitations under the License.
...
+      })
       .afterClosed()
-      .pipe(
-        filter(result => result.status === ModalDialogStatus.Ok)
-      )
-      .subscribe(result => {
-        alert(result.resolvedContent);
+      .pipe(filter((result) => result.status === ModalDialogStatus.Ok))
+      .subscribe((result: ConflictEditorModalResult) => {
+        if (result.status === ModalDialogStatus.Ok) {
```
pjmonks, line 125: You shouldn't need this `if` block, the `filter()` operator in the pipe means only `ModalDialogStatus.Ok` results get returned.

src/app/merge-diff/merge-diff-container/merge-diff-container.component.ts
```diff
@@ -207,11 +234,43 @@ export class MergeDiffContainerComponent implements OnInit {
...
     });
     this.changesList = new Array<FullMergeItem>();
+    this.changesList = Object.assign([], tempArray);
+  }
+
+  cancelAll() {
+    this.dialog
+      .openConfirmation({
+        data: {
+          message: 'Are you sure? this will remove all pending commits',
```
pjmonks, line 259: "This" should be capitalised

pjmonks (dismissed):
This is really good

pjmonks (dismissed):
Looking good, I like the split of the diff/render functionality into a shared service.

pjmonks (approved):
I think this new UI is looking excellent, well done!

PR #227, opened 2021-07-08
added check for folders and version folders and call new endpoint

gammonpeter:


pjmonks (changes requested):
The code looks good.

I'm not sure how your call to the new `getFolder` endpoint is working here, unless it's based on your working copy. If so, `mdm-resources` has been merged into `develop` now, so please run this to update the `package-lock.json` file and commit that also:

```shell
npm update @maurodatamapper/mdm-resources
```

pjmonks (approved):

PR #228, opened 2021-07-09
Feature/gh 181 Update icons for catalogue domain types

### LINKED ISSUE #181: Select appropriate icons for lower level catalogue items
image

Icons were chosen for Data Type and Data Class (see above) to have a consistent icon set when used with the model tree. The icons to be used though are up for debate. For instance, is the chosen Data Class icon above too similar to Versioned Folders?

@jamesrwelch , @olliefreeman, @gammonpeter  - any icon can be chosen to substitute, so long as they from the [free Font-Awesome list](https://fontawesome.com/v5.15/icons?d=gallery&p=2&m=free)

pjmonks:
Resolves #181

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "mdm-ui",
       "version": "6.4.0-SNAPSHOT",
       "hasInstallScript": true,
       "dependencies": {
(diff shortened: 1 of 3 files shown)


gammonpeter (approved):
Happy to approve, but I recommend we add a task to update the detailed components

PR #230, opened 2021-07-09
Feature/gh 160 Include "Add all Terms" option to multi-term selector

### LINKED ISSUE #160: 'Select all' option when adding terms to a codeset
When choosing terms from a terminology to add to a codeset, it would be nice if there was a 'select all' option.  There are similar components where we add types to a model, etc - would be nice to ensure we have that option everywhere.

pjmonks:
Resolves #160 

Update `mdm-multiple-terms-selector` component to include an "Add all Terms" option:

image

If ticked, then every term in the selected terminology will be added to the parent code set. Untick to revert back to the manual search/filter selection list to pick terms.

### DIFF
--- a/src/app/utility/multiple-terms-selector/multiple-terms-selector.component.html
+++ b/src/app/utility/multiple-terms-selector/multiple-terms-selector.component.html
@@ -17,74 +17,156 @@ limitations under the License.
 SPDX-License-Identifier: Apache-2.0
 -->
 <div class="mb-2">
-    <span class="block">Select a Terminology<sup class="warn">*</sup></span>
-    <mdm-select width="'100%'"
-                [acceptTypedInput]="false"
-                [valueType]="'static'"
-                [staticValues]="selectorSection.terminologies"
-                [idProperty]="'id'"
-                [displayProperty]="'label'"
-                [searchProperty]="'label'"
-                (selectEvent)="onTerminologySelect($event[0])">
-        <ng-template #lineContent let-item>
-            <div>{{item.label}}</div>
-        </ng-template>
-    </mdm-select>
+  <span class="block">Select a Terminology<sup class="warn">*</sup></span>
+  <mdm-select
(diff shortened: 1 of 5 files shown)


src/app/utility/multiple-terms-selector/multiple-terms-selector.component.html
```diff
@@ -17,74 +17,156 @@
...
+      <!-- <div style="height: 400px; "> -->
+      <mat-table
+        #table
+        [dataSource]="dataSource"
+        style="max-height: 100%; overflow: auto"
+      >
+        <ng-container matColumnDef="label">
+          <mat-header-row *matHeaderCellDef> Term </mat-header-row>
+          <mat-cell
+            style="
```
gammonpeter, line 71: Might be better placed in class

src/app/utility/multiple-terms-selector/multiple-terms-selector.component.ts
```diff
@@ -126,9 +129,11 @@ export class MultipleTermsSelectorComponent {
       this.currentRecord = 0;
     }
   };
+
   runTermSearch = () => {
```
gammonpeter, line 133: Might be good time to refactor this

gammonpeter (changes requested):
Not sure the checked box should default to true. I think it should be a positive choice to select all not one that's made for me.

src/app/utility/multiple-terms-selector/multiple-terms-selector.component.html
pjmonks: Done

src/app/utility/multiple-terms-selector/multiple-terms-selector.component.ts
pjmonks: Done

gammonpeter (approved):

PR #231, opened 2021-07-09
Added Reorder and fixed data model issue with redorder. Refactor clas…

gammonpeter:
…ses to match new standard

### DIFF
--- a/src/app/shared/data-classes-list/data-classes-list.component.ts
+++ b/src/app/shared/data-classes-list/data-classes-list.component.ts
@@ -16,7 +16,16 @@ limitations under the License.
 
 SPDX-License-Identifier: Apache-2.0
 */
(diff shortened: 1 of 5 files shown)


src/app/shared/data-classes-list/data-classes-list.component.ts
```diff
@@ -168,7 +208,7 @@ export class DataClassesListComponent implements AfterViewInit {
             });
          }
       });
-      const promise = new Promise((resolve, reject) => {
+      const promise = new Promise<void>((resolve, reject) => {
```
pjmonks, line 211: I think this wrapping of an observable into a promise should be refactored out. The actions done in the `promise.then()` should just be part of the `subscribe()` call.

pjmonks (changes requested):
If we can remove those old `Promise`-style wrappings of observables, the rest of the code looks good.

pjmonks (approved):

PR #238, opened 2021-07-15
Feature/mc 9563 DOI profile - Submission UI

pjmonks:
WIP of the new DOI profile submission UI.

First step is to refactor the existing default/other/profiles views embedded in every domain type view component - replication of code is causing issues with adding the new features required for DOI.

* Ported `ProfileBaseComponent` to a new shared component called `mdm-profile-data-view`.
* Shared component does the same job as before - displaying correct profile view, edits, etc - but does not require every domain type view to integrate so tightly, can now act as a child component.

Currently only the `DataModelComponent` view has been modified to display this new shared profile view as a test case to prove it works.


gammonpeter (approved):
Good Work

pjmonks:
PR is now updated with all parts, hoping that backend endpoints are completed.

jamesrwelch (approved):
Looks good to me - nice one!

PR #247, opened 2021-07-23
Feature/mc 9541 Updated merge UI v2 with backend endpoints

pjmonks:
The 2nd version of the merge UI has been updated with the real endpoints for diffs and patching, currently only for Data Models.

* Tested on different types of value changes - any known issues have been reported to `mdm-core`
* Added new `NumberConflictEditorComponent` to handle number conflicts - this is primarily for multiplicity values that change between source and target. Nearly everything else is a string.
* Minor UI tweaks:
    * Add counters to the tabs for "Changes" and "Committing" lists
    * Updated the Check In dialog to include more information
    * Updated the path names parsing as the format has changed on the backend.
    * Re-organised the menu items in the detail views, "Merge" actions are no longer in a sub-menu. This is so that the "Show merge graph" is more prominent to a user to help them track branches better


pjmonks:
Updated this PR to include [MC-9552](https://metadatacatalogue.myjetbrains.com/youtrack/issue/MC-9552) - updating the merge graph view to also support Versioned Folders:

* Refactored to clean up the code in the dialog window
* Updated to support all mergable domain types - Data Models, Versioned Folders, Terminologies, Code Sets and Reference Data Models (though UI only currently uses the first two).

jamesrwelch (approved):
Looks like a lot of work has gone into this - nice!

PR #248, opened 2021-07-26
Feature/mc 9564 DOI Resolution landing page

pjmonks:
New UI landing page at URL route `/#/doi/{doi}` to automatically resolve a catalogue item given a DOI and then redirect to that catalogue item.


jamesrwelch (approved):
This all looks fine.  The styling in doi-redirect.component.scss looks like it could be more generic... but happy to accept this now.

PR #252, opened 2021-07-27
Feature/gh 250 Feedback changes to new merge conflict editor

### LINKED ISSUE #250: Change conflict editing experience of the new Merge interface
Feedback from users regarding the new conflict resolution editor:

- [x]  Do not allow this warning to appear - before resolving conflicts, all conflicts must be resolved.

image

- [x] Add toolbar buttons to "Add all Source" or "Add all Target". This would then copy all the values from one side into the resolved editor.

- [x] The "Cancel All" toolbar button in the "Committing" list does not remove the resolution tags in the path names when in the "Changes" list. A single "Cancel commit" action does do this though.

pjmonks:
Resolves #250 

1. Removes user warning if conflicts are still found in resolved text. Must now remove every conflict before accepting.
2. Add "Take all" buttons to source/target views to for copying every change into resolved for fixing conflicts.
3. Fixed bug in cancelling all commits where branch name was still present.

### DIFF
--- a/src/app/merge-diff/conflict-editor/conflict-editor-modal/conflict-editor-modal.component.html
+++ b/src/app/merge-diff/conflict-editor/conflict-editor-modal/conflict-editor-modal.component.html
@@ -84,28 +84,18 @@ SPDX-License-Identifier: Apache-2.0
       No
     </button>
   </div>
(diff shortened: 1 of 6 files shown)


jamesrwelch (approved):
Looks spiffing!  Good work!

PR #253, opened 2021-07-27
#251 Change layout of login dialog

### LINKED ISSUE #251: External provider login buttons should appear next to login form
image

Currently the external login provider buttons appear below the form, but they should probably be to the right of the existing username/password form.

If the width of the dialog/page is too narrow, the buttons should be responsive and flex down into a column again.

pjmonks:
Resolves #251 

* Email/password form is now on left and OpenID Connect buttons are on right (if available)
* Mobile view will respond to stacking controls in a column

Desktop view:

image

Mobile view:

image


jamesrwelch (approved):
Beautiful!

PR #256, opened 2021-07-28
#245 Allow user groups to be added to finalised models

### LINKED ISSUE #245: Cannot change permissions on a finalised model
**Describe the bug**
The 'Add Group' button does not appear on the permissions modal window when viewing a finalised model.
I believe any changes to the 'publicly readable' and 'readable by authenticated users' are also not persisted.

**Expected Behaviour**
You should be able to change the permissions of a finalised model.

pjmonks:
Fixes #245


jamesrwelch:
This looks simple enough, but for some reason a test is failing.  Just a question of re-running?

pjmonks:
@jamesrwelch It's the Jenkins build being funny, it does that sometimes. Nothing is wrong with the branch, all passes now.

jamesrwelch (approved):
Cool

PR #260, opened 2021-07-30
MC-9563 Update availableActions with corrected names

pjmonks:
Change required to use the new `finalisedEditActions` and `finalisedReadActions` sent from the backend. This is required to get the DOI submission options to appear.


jamesrwelch (approved):
Makes sense

PR #262, opened 2021-07-30
#261 Fix string diff comparison if either of the compare strings are null

### LINKED ISSUE #261: Merge diff comparison when one side is `undefined` throws console error
**Describe the bug**
If comparing a field in a merge diff where one side has an `undefined` value (e.g. the value was not set yet), then trying to view the comparison throws a JS error to the console and the component views do not refresh correctly.

**To Reproduce**
1. Create a Data Model for the "main" branch
2. Next, branch of it to create a "test" branch
3. In the "test" branch, update the description field and save
4. Now merge "test" into "main"
5. Find the `@description` field change, make sure it is in the "Changes" list, and select it to see the comparison below.

**Expected behavior**
One branch with the value should be displayed, while the `undefined` side should have some default text display to show it is not set.

**Screenshots**
image

pjmonks:
Resolves #261

### DIFF
--- a/src/app/merge-diff/merge-comparsion/merge-comparsion.component.ts
+++ b/src/app/merge-diff/merge-comparsion/merge-comparsion.component.ts
@@ -192,8 +192,8 @@ export class MergeComparisonComponent implements OnInit, OnChanges {
   }
 
   private getComparisonDisplayText(value: string, other: string) {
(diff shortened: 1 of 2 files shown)


jamesrwelch (approved):
Seems reasonable

PR #285, opened 2021-08-18
Update profiles to support changes on backend

### LINKED ISSUE #284: Update profiles to support changes to validation, finalisation and read-only fields
`mdm-core` now has changes to the profiles endpoints to support the following:

1. Validation now returns a `422` if errors are found, with a map of error messages and field names. This has changed from original, which returned the full profile again with error messages embedded inside the fields.
2. Profiles can now have derived and read-only fields
3. Profiles can now support being edited after a model is finalised

Make sure the UI can now support these features.

- [x] Display read-only fields if the profile includes them
- [x] Update validation to use updated response
- [x] Fields which include `allowedValues` should be presented as a drop-down
- [x] Required fields should use the `minMultipicity` field
- [x] Test that post-finalise editing is supported

pjmonks:
Fixes #284

* Uses correct property names and data types returned from profiles - names have been made more consistent
* Supports uneditable fields
* Validation endpoint now returns a collection of errors when a `422` is returned, so UI has been updated to reflect this
* Fixed required fields by looking at `minMultiplicity` field property

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,7 +24,7 @@
         "@bpmn-io/dmn-migrate": "^0.4.3",
         "@ctrl/ngx-codemirror": "4.0.1",
         "@fortawesome/fontawesome-free": "5.15.1",
-        "@maurodatamapper/mdm-resources": "git+https://github.com/MauroDataMapper/mdm-resources.git#develop",
+        "@maurodatamapper/mdm-resources": "git+https://github.com/MauroDataMapper/mdm-resources.git#feature/profile-updates",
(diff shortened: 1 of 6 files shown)


gammonpeter (dismissed):

olliefreeman:
> also would it be possible to have uneditable fields look "open" rather than greyed out like the derived fields. (note i can see that the `---` gets converted to a line break ... this is custom to the doi profile, so i will remove it now.
> 
> Screenshot 2021-08-19 at 09 41 29

gammonpeter (approved):

PR #311, opened 2021-09-01
Feature/MC-9483 Part 1

atpw25:
This pull request contain majority of the changes for MC-9483 except for the last feature (Create Ref DataType using imported DataClass). The last feature require changes to `mdm-core` that are still in development.

### DIFF
--- a/src/app/shared/data-classes-list/data-classes-list.component.ts
+++ b/src/app/shared/data-classes-list/data-classes-list.component.ts
@@ -317,16 +317,7 @@ export class DataClassesListComponent implements AfterViewInit {
   }
 
   bulkDelete() {
-    const dataElementIdLst = [];
-    this.dataClassRecords.forEach((record) => {
-      if (record.checked) {
-        dataElementIdLst.push({
-          id: record.id,
-          domainType: record.domainType
-        });
-      }
-    });
-
+    const dataElementIdLst = this.dataClassRecords.filter(record => record.checked === true);
     this.dialog
       .open(BulkDeleteModalComponent, {
         data: {
--- a/src/app/shared/data-classes-list/data-classes-list.component.ts
+++ b/src/app/shared/data-classes-list/data-classes-list.component.ts
@@ -82,6 +82,7 @@ export class DataClassesListComponent implements AfterViewInit {
     private gridService: GridService,
     private messageHandler: MessageHandlerService
   ) {}
+
   ngAfterViewInit() {
     if (this.isEditable && !this.parentDataModel.finalised) {
       this.displayedColumns = ['name', 'description', 'label', 'checkbox'];
@@ -112,7 +113,11 @@ export class DataClassesListComponent implements AfterViewInit {
           );
         }),
         map((data: any) => {
-          this.totalDataClassCount = data.body.count;
+          if (this.parentDataClass.extendsDataClasses) {
+            this.totalDataClassCount = parseInt(data.body.count, 10) + (this.parentDataClass.extendsDataClasses.length as number);
+          } else {
+            this.totalDataClassCount = data.body.count;
+          }
           this.isLoadingResults = false;
           return data.body.items;
         }),
@@ -122,7 +127,15 @@ export class DataClassesListComponent implements AfterViewInit {
         })
       )
       .subscribe((data) => {
-        this.dataClassRecords = data;
+        if (this.parentDataClass.extendsDataClasses) {
+          const extendedDC = this.parentDataClass.extendsDataClasses.map(dc => {
+            dc['extended'] = true;
+            return dc;
+          });
+          this.dataClassRecords = [...data, ...extendedDC];
+        } else {
+          this.dataClassRecords = data;
+        }
       });
   }
 
--- a/src/app/shared/data-classes-list/data-classes-list.component.ts
+++ b/src/app/shared/data-classes-list/data-classes-list.component.ts
@@ -38,7 +38,7 @@ import { BulkEditModalComponent } from '@mdm/modals/bulk-edit-modal/bulk-edit-mo
 import { BulkDeleteModalComponent } from '@mdm/modals/bulk-delete-modal/bulk-delete-modal.component';
 import { GridService } from '@mdm/services/grid.service';
 import { CdkDragDrop, moveItemInArray } from '@angular/cdk/drag-drop';
-import { DataClass } from '@maurodatamapper/mdm-resources';
+import { DataClass, DataClassDetail, DataClassIndexResponse, DataModelDetail } from '@maurodatamapper/mdm-resources';
 import { MessageHandlerService } from '@mdm/services';
 
 @Component({
@@ -51,19 +51,18 @@ export class DataClassesListComponent implements AfterViewInit {
   @ViewChild(MatSort, { static: false }) sort: MatSort;
   @ViewChild(MdmPaginatorComponent, { static: true })
   paginator: MdmPaginatorComponent;
-  @ViewChild(MatTable, { static: false }) table: MatTable<any>;
+  @ViewChild(MatTable, { static: false }) table: MatTable<DataClassDetail>;
 
-  @Input() parentDataModel: any;
-  @Input() grandParentDataClass: any;
-  @Input() parentDataClass: any;
-  @Input() loadingData: any;
-  @Input() isEditable: any;
+  @Input() parentDataModel: DataModelDetail;
+  @Input() grandParentDataClass: DataClassDetail;
+  @Input() parentDataClass: DataClassDetail;
+  @Input() isEditable: boolean;
   checkAllCheckbox = false;
 
   processing: boolean;
   failCount: number;
   total: number;
-  dataClassRecords: any[] = [];
+  dataClassRecords: DataClassDetail[] = [];
   hideFilters = true;
   displayedColumns: string[];
   loading: boolean;
@@ -112,9 +111,9 @@ export class DataClassesListComponent implements AfterViewInit {
             this.filter
           );
         }),
-        map((data: any) => {
+        map((data: DataClassIndexResponse) => {
           if (this.parentDataClass.extendsDataClasses) {
-            this.totalDataClassCount = parseInt(data.body.count, 10) + (this.parentDataClass.extendsDataClasses.length as number);
+            this.totalDataClassCount = data.body.count + (this.parentDataClass.extendsDataClasses.length as number);
           } else {
             this.totalDataClassCount = data.body.count;
           }
@@ -139,7 +138,7 @@ export class DataClassesListComponent implements AfterViewInit {
       });
   }
 
-  openEdit(dataClass: DataClass) {
+  openEdit(dataClass: DataClassDetail) {
     if (!dataClass || (dataClass && !dataClass.id)) {
       return '';
     }
--- a/src/app/shared/data-classes-list/data-classes-list.component.ts
+++ b/src/app/shared/data-classes-list/data-classes-list.component.ts
@@ -317,7 +317,7 @@ export class DataClassesListComponent implements AfterViewInit {
   }
 
   bulkDelete() {
-    const dataElementIdLst = this.dataClassRecords.filter(record => record.checked === true);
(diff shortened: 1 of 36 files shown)


pjmonks:
I've been having a lengthy discussion with @jamesrwelch and @olliefreeman about the UX of this feature as designed and believe there could be some improvements made (through no fault of your own).

# UX Issues

The main issue I see here (imagining I'm a new user who has never seen this before) is that the word "import" is used in multiple ways in the UI now and I don't believe the word is appropriate for a user in this use case. "Import" to me suggests that data will be taken from one source (e.g. external, file upload etc) and then _copied_ into the target. The backend uses this definition throughout, yet when I initially looked at the UI during testing, I got confused and wondered why the data I had "imported" was actually belonging to something else.

For comparison, there is existing functionality in MDM which imports data models, terminologies, etc via something like JSON, XML etc. In these cases, the data comes from a source and is _copied_ into the catalogue. When compared to this feature of import/extending, why isn't the same happening?

# Improvements

Luckily, I only have a few suggestions to make which I think would improve the user experience without causing a lot of rework. As it stands, the backend can keep referring to this concept as "importing" (as explained to me, this is the correct terminology for this), yet the UI can be re-worded to use simpler definitions for a user who does not need to understand the technical details of what is happening in the background.

## Consolidate tree right-click menu

image

Currently, right-clicking a tree node produces two sub-menus, yet each menu option from each always navigates you to the same "New..." wizard view, Also, now matter which menu is chosen, the same default options in the wizard are picked. This is duplication, so:

1. Remove the "Create..." and "Import/Extend..." menus.
2. Instead, replace with the "Add" menu, which then displays a sub-menu of item types e.g. Data Class, Data Element, Data Type (as appropriate)

Using the word "Add..." can imply all the possible operations allowed - create from scratch, copy/clone from another source, or reference from somewhere else.

## Wizard options

image

For the wizard options made available, rename "Import X from..." to "Use X from...". "Use" does not then get mixed up with "Copy".

I think "Extend with Data Class(es) from..." could remain as it is for now.

## Source badges

image

Finally, at the points where it is made clear that a catalogue item actually comes from somewhere else, change the word "Imported" to "Reference" - to highlight that this item is more like a reference, or link, to the real item somewhere else.

Alternatively, you could instead use "Source: xxx", where "xxx" is the actual name of the model it came from (I believe the identifier of the model is provided in the JSON response alongside the `imported` boolean flag). This would make it much cleared that the item is from somewhere else, though it depends if there is space available for long model names.

# Review

If you could look at doing these improvements @atpw25 please, I will continue to review this PR in the meantime.

src/app/shared/data-classes-list/data-classes-list.component.ts
```diff
@@ -112,7 +113,11 @@ export class DataClassesListComponent implements AfterViewInit {
           );
         }),
         map((data: any) => {
-          this.totalDataClassCount = data.body.count;
+          if (this.parentDataClass.extendsDataClasses) {
```
pjmonks, line 116: I've spotted that there is legacy code in the `mdm-data-classes-list` component, where `any` is used too often. I think, seeing as you have edited this component last, this would be a good opportunity to remove `any` and replace with defined types for better compiler checks. For instance:

1. `parentDataModel` should probably be a `DataModelDetail`
2. `parentDataClass` should probably be a `DataClassDetail`
3. `isEditable` should be a `boolean`
4. And so on...

Also:

1. `loadingData` doesn't actually seem to be used in this component. If that is true, please remove it.

src/app/shared/data-classes-list/data-classes-list.component.ts
```diff
@@ -304,16 +317,7 @@ export class DataClassesListComponent implements AfterViewInit {
...
-    this.dataClassRecords.forEach((record) => {
-      if (record.checked) {
-        dataElementIdLst.push({
-          id: record.id,
-          domainType: record.domainType
-        });
-      }
-    });
-
+    const dataElementIdLst = this.dataClassRecords.filter(record => record.checked === true);
```
pjmonks, line 320: Nice consolidation of code here. For the `filter()` predicate, you may even reduce it further to `filter(record => record.checked)`

src/app/shared/data-elements-list/data-elements-list.component.ts
```diff
@@ -295,15 +295,7 @@ export class DataElementsListComponent implements AfterViewInit {
...
-    const dataElementIdLst = [];
-    this.dataElementRecords.forEach((record) => {
-      if (record.checked) {
-        dataElementIdLst.push({
-          id: record.id,
-          domainType: record.domainType
-        });
-      }
-    });
+    const dataElementIdLst = this.dataElementRecords.filter(record => record.checked === true);
```
pjmonks, line 298: Again, I like how this is reduced, could also remove the equality check as well if `checked` is a boolean.

src/app/shared/element-child-data-classes-list/element-child-data-classes-list.component.html
```diff
@@ -55,6 +55,7 @@
           <td mat-cell *matCellDef="let record" style="overflow-wrap: break-word; word-break: break-word;">
             <div>
                 <mdm-element-link [element]="record" [parentDataModel]="{id:record.dataModel}"></mdm-element-link>
+                <div *ngIf="record.imported"><small><em><span class="badge badge-success">imported</span></em></small></div>
```
pjmonks, line 58: See comments in PR about UX improvements.

src/app/shared/element-owned-data-type-list/element-owned-data-type-list.component.html
```diff
@@ -61,6 +61,7 @@ <h4 class="marginless">
             <td mat-cell *matCellDef="let record" style="max-width: 25%;width: 25%;text-align: left; overflow-wrap: break-word; word-break: break-word;">
                 <div>
                     <mdm-element-link [element]="record" [parentDataModel]="{id:record.model}"></mdm-element-link>
+                    <div *ngIf="record.imported"><small><em><span class="badge badge-success">imported</span></em></small></div>
```
pjmonks, line 64: See comments in PR about UX improvements

src/app/wizards/dataClass/data-class-step1/data-class-step1.component.html
```diff
@@ -23,18 +23,18 @@
...
-    <mat-radio-group name="createType" [(ngModel)]="model.createType" (click)="selectCreateType('new')">
-        <mat-radio-button [checked]="model.createType === 'new'" value="new">Create a New Data Class</mat-radio-button>
-    </mat-radio-group>
 
-    <mat-radio-group name="createTypeCopy" [(ngModel)]="model.createType" (click)="selectCreateType('copy')">
-        <mat-radio-button [checked]="model.createType === 'copy'" value="copy">Copy Data Class(s) from...</mat-radio-button>
+    <mat-radio-group name="createType" [(ngModel)]="model.createType" (change)="validate()">
+      <mat-radio-button value="new">Create a New Data Class</mat-radio-button>
+      <mat-radio-button value="copy">Copy Data Class(s) from...</mat-radio-button>
+      <mat-radio-button value="import">Import Data Class(s) from...</mat-radio-button>
```
pjmonks, line 30: See comments in PR about UX improvements

src/app/wizards/dataClass/data-class-step2/data-class-step2.component.ts
```diff
@@ -190,7 +191,14 @@ export class DataClassStep2Component
 
       invalid = this.myForm.invalid;
     }
-    if (this.model.createType === 'copy') {
+    if (['copy', 'import', 'extend'].includes(this.model.createType)) {
```
pjmonks, line 194: I think `createType` could benefit from using a union string type here to ensure the string is always correct, e.g.:

```ts
export type CreateType = 'copy' | 'import' | 'extend';
```

src/app/wizards/dataElement/data-element-step1/data-element-step1.component.html
```diff
@@ -20,17 +20,16 @@
     <span>How do you want to add a new Data Element(s) to </span>
     <mdm-element-link [newWindow]="true" [element]="model.parent"></mdm-element-link>
     <span>?</span>
-    <mat-radio-group name="createType" [(ngModel)]="model.createType" (click)="selectCreateType('new')">
+    <mat-radio-group name="createType" [(ngModel)]="model.createType" (change)="validate()">
         <mat-radio-button value="new">Create a New Data Element</mat-radio-button>
-    </mat-radio-group>
-    <mat-radio-group name="createType" [(ngModel)]="model.createType" (click)="selectCreateType('copy')">
         <mat-radio-button value="copy">Copy Data Element(s) from...</mat-radio-button>
+        <mat-radio-button value="import">Import Data Element(s) from...</mat-radio-button>
```
pjmonks, line 26: See comments in PR about UX improvements

src/app/wizards/dataElement/data-element-step2/data-element-step2.component.ts
```diff
@@ -244,7 +244,7 @@ export class DataElementStep2Component implements OnInit, AfterViewInit, OnDestr
   validate = (newValue?) => {
     let invalid = false;
     this.step.invalid = false;
-    if (newValue && this.model.createType === 'new') {
+    if (newValue && ['copy', 'import'].includes(this.model.createType)) {
```
pjmonks, line 247: See previous comment about defining a union string type for these constants.

src/app/wizards/dataType/data-type-step1/data-type-step1.component.html
```diff
@@ -19,17 +19,16 @@
...
 	<mdm-element-link [parentDataModel]="{id:model.parent}" [newWindow]="true" [element]="model.parent"></mdm-element-link>
-	<mat-radio-group name="createType" [(ngModel)]="model.createType" (click)="selectCreateType('new')">
-		<mat-radio-button [checked]="model.createType === 'new'" value="new">Create a New Data Type</mat-radio-button>
-	</mat-radio-group>
-	<mat-radio-group name="createType" [(ngModel)]="model.createType" (click)="selectCreateType('copy')">
-		<mat-radio-button [checked]="model.createType === 'copy'" value="copy">Copy Data Type(s) from...</mat-radio-button>
+	<mat-radio-group name="createType" [(ngModel)]="model.createType" (change)="validate()">
+		<mat-radio-button value="new">Create a New Data Type</mat-radio-button>
+		<mat-radio-button value="copy">Copy Data Type(s) from...</mat-radio-button>
+    <mat-radio-button value="import">Import Data Type(s) from...</mat-radio-button>
```
pjmonks, line 25: See comments in PR about UX improvements

src/app/wizards/dataType/data-type-step2/data-type-step2.component.ts
```diff
@@ -139,9 +139,9 @@ export class DataTypeStep2Component implements OnInit, AfterViewInit, OnDestroy
       this.dataSourceDataTypes.sort = this.sort.toArray()[0];
 
       // Selected Data Types table
-      this.dataSourceSelectedDataTypes.sort = this.sort.toArray()[1];
-      this.sort.toArray()[1].sortChange.subscribe(() => (this.paginator.toArray()[1].pageIndex = 0));
-      this.dataSourceSelectedDataTypes.paginator = this.paginator.toArray()[1];
+      // this.dataSourceSelectedDataTypes.sort = this.sort.toArray()[1];
```
pjmonks, line 142: Remove commented out lines if no longer necessary.

src/app/wizards/dataType/data-type-step2/data-type-step2.component.ts
```diff
@@ -267,7 +267,7 @@ export class DataTypeStep2Component implements OnInit, AfterViewInit, OnDestroy
       }
       invalid = true;
     }
-    if (this.model.createType === 'copy') {
+    if (['copy', 'import'].includes(this.model.createType)) {
```
pjmonks, line 270: See previous comment about a union string type for these constants.

atpw25:
I think I have covered the changes requested with the latest changes. Please review.

pjmonks (approved):
I'm happy with all these changes, and the UX improvements I think look better. Thanks @atpw25

PR #323, opened 2021-09-22
#322 Fix OpenID Connect redirect issue with URL subdirectories

### LINKED ISSUE #322: Admin OIDC ui doesn't handle non-root deployed instances
**The bug**
The OIDC connect redirect URL does not include the subfolder, such as https://mydomain.com/nhsd-test/

The result is that the apache server gives a generic 404 error

pjmonks:
If the server uses a subdirectory as well as the domain, the original redirect was failing because it was not accounted for.

* Update redirect static page to correctly build the URL for redirection to the angular route
* Update OIDC provider page to display correct redirect URL info

This can be tested locally by passing the `--base-ref` parameter to `ng serve`, e.g. `ng serve --base-href /nhsd-test/`

Fixes #322


pjmonks:
Closing this PR - carried out a hotfix onto `main` branch instead, see https://github.com/MauroDataMapper/mdm-ui/releases/tag/6.6.1.

PR #377, opened 2021-10-28
Corrected property name

gammonpeter:


jamesrwelch (approved):
Nice - this works.  I'll create a new ticket for improving the messages when there are no results available after filtering

PR #379, opened 2021-10-29
Method was using incorrect id

gammonpeter:


jamesrwelch (commented):
Tested this on a model, and it now works.  But getting an odd error when linking to a data class which may be related or may be a separate issue.  Links are not generated for a data type at all, which is almost certainly a separate issue.

Issue #382, opened 2021-10-28
Advanced Search does not work for public user


jelenadreskai:
1. Do not log in to Mauro. Look around using unauthenticated user.
2. Click on the Search
3. Click on Advanced option

Expected results:
Advanced search form shows up

Current behavior:
User sent to the main MDM page

--
MDM Release
UI: 6.6.0
CORE: 4.10.0

jamesrwelch:
If you're able to get a PR in for this one by Wednesday, we'll be able to include it in the next release :)

GBishop-PHSA:
Clicking the advanced search link when not logged in was resulting in a request of:
/api/tree/folders?queryStringParams=[object Object]
The queryStringParams should be expanded before sending the request, so it sends as:
/api/tree/folders?includeDocumentSuperseded=true&includeModelSuperseded=true&includeDeleted=true

GBishop-PHSA:
@jamesrwelch The pull request is ready for review

PR #385, opened 2021-11-01
GH-382 Fix to load tree with default filters

GBishop-PHSA:
Clicking the advanced search link when not logged in was resulting in an incorrect request call to load tree endpoint.


GBishop-PHSA:
@gammonpeter @jamesrwelch The pull request is ready for review.

jamesrwelch (approved):
Lovely work - thank you!

PR #389, opened 2021-11-02
Feature/gh 327

gammonpeter:


jamesrwelch (changes requested):
In this branch, when I submit a new comment, I get a lot of:
```
TypeError: undefined is not an object (evaluating 'record_r9.createdByUser.name')
```
And no success - the comment is created when I refresh the page, but the 'working' and auto-refresh never happens.

jamesrwelch (approved):
Yep, lovely work

PR #395, opened 2021-11-12
updated so clear fires change event

gammonpeter:


jamesrwelch:
Now getting: 
```
Schema validation failed with the following errors:
  Data path "/polyfills" must be string.
```
On this branch... what have I done wrong?

pjmonks:
I have tested this and have no issue with the UI now. I did find a related error in `mdm-core` though which prevents you from filtering searches when setting the "Last Updated - To" date filter - this is recorded in MauroDataMapper/mdm-core#230. I think that is separate though, since changing the filter field does trigger the search again, which is what this PR is meant to fix.

pjmonks (approved):

PR #416, opened 2022-01-07
#412 Fix refresh of federated models after subscribing to an external model

### LINKED ISSUE #412: Subscribed models don't refresh after subscribing to a model
**Describe the bug**
After successfully subscribing to a published model, and then selecting other published models under External catalogues in the navigation tree, the first published model stays displayed and the selected subscribed model isn't displayed. A workaround is to refresh the page or navigate to something that isn't a subscribed model.

**To Reproduce**
1. Select a published model
2. Click Subscribe and successfully subscribe
3. Immediately click on a different published model; the first published model remains displayed

**Expected behavior**
Clicking on a published model should navigate to it after subscribing to another.

**Desktop:**
 - OS: macOS
 - Browser: Chrome
 - Version: 96

**Additional context**
Unrelated to the changes from mc-9735.

pjmonks:
After subscribing, the router would navigate back to the same page and pass the loaded model as a route parameter. Because no other navigation sites pass this parameter, it hung around for subsequent calls causing the page to not to load the correct selection from the model tree.

Fixes #412

### DIFF
--- a/src/app/subscribed-catalogues/federated-data-model-main/federated-data-model-main.component.ts
+++ b/src/app/subscribed-catalogues/federated-data-model-main/federated-data-model-main.component.ts
@@ -102,8 +102,7 @@ export class FederatedDataModelMainComponent extends BaseComponent implements On
       'federateddatamodel',
       {
         parentId: this.catalogueId,
-        id: this.modelId,
-        dataModel: this.dataModel
+        id: this.modelId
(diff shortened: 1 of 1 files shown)


joe-crawford (commented):
I've tested and this works for me.

joe-crawford (approved):

PR #425, opened 2022-01-11
New page to create a blank Terminology

### LINKED ISSUE #406: Create Terminology from UI
As part of the Terminology editing functionality (MC-9606), we need also to be able to create a new terminology from the UI. 
Screenshot 2021-12-15 at 09 00 47
In the same way that we currently create a data model or a codeset.

pjmonks:
Resolves #406 

Created new page to create a new, blank Terminology - accessible via the right-click menu of the model tree:

image

image

Requires `mdm-resources` to be up to date with required endpoint added, see MauroDataMapper/mdm-resources#41


jamesrwelch (approved):
This works very nicely.  At some point we should unify all the model creation pages

PR #431, opened 2022-01-14
Bulk edit of profiles for Data Elements

### LINKED ISSUE #233: Enhancement suggestion: bulk view and edit of profile elements
When a user wishes to view or edit the value of a profile property against a Data Element, they can take the following path:

- Click on Data Model
- Click on Schema tab and click on relevant Data Class
- Click on the Elements tab
- Click on the relevant Data Element
- Select the relevant profile from a dropdown
- View and edit the profile values for the selected Data Element

It would be good if users could:
1. get to the profile values with fewer clicks
2. see and edit the profile values for multiple data elements in one place.

An example of (2) would be a tabular edit/display with Data Elements as rows and profile attributes as columns. How to deal with several Data Elements having different profiles would need some more thought.

pjmonks:
Resolves #233

First iteration of the bulk editor for updating multiple elements (via profiles) at once. Initial implementation only allows editing of Data Elements for now.

image

image

image

Current feature set:

* Applies to Data Models only, by selecting Data Elements under a model
* Multiple profiles can be chosen, with a tab created for each one
* Profile sections and fields are grouped as column groups with sub-columns
* Multiple custom cell editors are supported at the moment:
    * Checkbox (boolean)
    * Date
    * Select (enumeration of allowed values)
 * Supports validation of profiles and saving with or without valid data

Requires the PR MauroDataMapper/mdm-resources#42 to be approved first, as that contains endpoints required by these components.


jamesrwelch (approved):
Brilliant!  First iteration of a popular feature!

PR #433, opened 2022-01-17
#430 Display all classifiers in selection list

### LINKED ISSUE #430: Classification list limited to 10 items [limitation]
Open an existing datamodel
Select Default Profile
Click on Edit Default Profile
Select Classifiers drop down list
The list contains only 10 classifiers. 

Note: The same behavior is in the Advanced Search, where list of classifiers shows up. 

Expected: 
All public classifiers should be shown in the drop down list.

UI: 6.6.0
CORE: 4.10.0

@jamesrwelch @olliefreeman @pjmonks 
Our solution was to increase the limit to 100 (because we do not expect more than 100 classifiers). This was quick and easy. If you want this code change, assign it to Gail.

But perhaps, a different UI control could be introduced to type and search for classifier. Or perhaps make this configurable.

pjmonks:
Resolves #430

### DIFF
--- a/src/app/search/advanced-search-bar/advanced-search-bar.component.ts
+++ b/src/app/search/advanced-search-bar/advanced-search-bar.component.ts
@@ -119,7 +119,7 @@ export class AdvancedSearchBarComponent implements OnInit {
     this.advancedSearch = false;
 
     this.resources.classifier
-      .list()
+      .list({ all: true })
       .subscribe((result: ClassifierIndexResponse) => {
         this.classifications = result.body.items;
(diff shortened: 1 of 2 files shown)


jamesrwelch (commented):
This resolves the issue when editing the default profile.  Issue #430 also mentions the same issue in the advanced search - can we make the same change there?

jamesrwelch:
I've made the same change there.  I can't see any other occurrences

jamesrwelch (approved):
Happy with this

Issue #413, opened 2022-01-06
Data Class: Sort by Element Name is not working


jelenadreskai:
**Describe the bug**
Data Class: Sort by Element Name is not working

**To Reproduce**
1. Open Data Model
2. Select one of the data classes
3. Click on the Elements tab
4. Click on the sort by element name

Result:
> Elements do not get sorted by name. Occasionally they get sorted, but very rare, and behavior is inconsistent.

Expected results:
> Elements are sorted by the element name

**MDM Version:**
-MDM 6.6.0/4.10

jelenadreskai:
@olliefreeman please assign to @GBishop-PHSA she has a solution.

PR #434, opened 2022-01-18
gh-413 Fix to Data Elements sort

GBishop-PHSA:
gh-413 Removed hard-coded sort by idx for Data Elements, as it was overriding the sortBy parameter in the method.

### DIFF
--- a/src/app/shared/data-elements-list/data-elements-list.component.ts
+++ b/src/app/shared/data-elements-list/data-elements-list.component.ts
@@ -116,9 +116,9 @@ export class DataElementsListComponent implements AfterViewInit {
       )
       .subscribe((data) => {
         this.dataElementRecords = data;
+        this.isLoadingResults = false;
       });
-
-    this.changeRef.detectChanges();
+      this.changeRef.detectChanges();
   }
 
   openEdit(dataClass: DataClass) {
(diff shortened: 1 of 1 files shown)


src/app/shared/data-elements-list/data-elements-list.component.ts
```diff
@@ -116,9 +116,9 @@ export class DataElementsListComponent implements AfterViewInit {
       )
       .subscribe((data) => {
         this.dataElementRecords = data;
+        this.isLoadingResults = false;
+        this.changeRef.detectChanges();
       });
-
-    this.changeRef.detectChanges();
```
pjmonks, line 121: The build server has detected a test that failed due to this line being removed. I've been playing with this locally and, although the sorting does work correctly when run in the UI, the `data-elements-list.component.spec.ts` file throws this error:

```
NG0100: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'true'. Current value: 'false'.. Find more at https://angular.io/errors/NG0100
```

I'm not exactly sure why this would happen, but I've been experimenting and think if you made these changes then both the test will pass _and_ the UI will sort the table correctly:

1. Add back the `changeRef.detectChanges()` call at the end of this method (_outside_ the `subscribe()` method).
2. Remove the `changeRef.detectChanges()` from inside the `subscribe()` callback.

Please could you action that?

pjmonks (changes requested):
The rest looks fine apart from that test failing I mentioned.

GBishop-PHSA:
@pjmonks I reverted a change to the code that had caused the test to fail. Ready for re-review.

pjmonks (approved):
All looks good @GBishop-PHSA , thanks.

PR #435, opened 2022-01-19
MC-9766 - Reorganise the repo to support multiple Angular projects

pjmonks:
The source code for the `mdm-ui` application has been reorganised to support [multiple Angular projects within the same workspace](https://angular.io/guide/file-structure#multiple-projects). This will prepare the repo to contain further UI applications/projects.

image

Changes made:

* Original files from the `src` folder have been moved to `projects/mdm-ui/src`. Files have been moved using `git mv` command to only rename files and maintain original commit history.
* All configuration files have been updated to support the change in file locations, particularly the `angular.json` file which is now responsible for all Angular projects inside this workspace.
* Verified that existing `npm` scripts still work. The default workspace project is still `mdm-ui`, so existing commands like `npm start` will still default to serving `mdm-ui`. Other scripts like `test`, `eslint`, `dist` etc have been updated to work with the new structure.
* The `Jenkinsfile` has been updated to ensure that any file references now point to new locations.

@olliefreeman - I have asked for your review also specifically to check the `Jenkinsfile`. I've seen that a Jenkins branch build passes still, and because `npm run dist` is used and still outputs the build output to `dist/mdm-ui-<version>,tgz` then there should not be any major changes with deployments.

### DIFF
--- a/projects/mdm-ui/src/app/bulk-edit/bulk-edit-editor/bulk-edit-editor.component.scss
+++ b/projects/mdm-ui/src/app/bulk-edit/bulk-edit-editor/bulk-edit-editor.component.scss
@@ -16,7 +16,7 @@ limitations under the License.
 
 SPDX-License-Identifier: Apache-2.0
 */
-@import "style/abstracts/variables";
+@import "../../../style/abstracts/variables";
 
 .mdm-bulk-edit-editor {
   &__validation {


pjmonks:
Marking this as a draft PR for now. Due to the way some other planned work has panned out, this re-organisation of `mdm-ui` may not be strictly necessary, or at least not urgent for now. Will keep open though in case it is needed.

PR #436, opened 2022-01-20
#399 Sort related terms by code explicitly

### LINKED ISSUE #399: When viewing a term, list of related terms should be ordered by code
When looking at related terms, they should be ordered by code - they're currently in a random order.  May require an issue raised against core to add ordering, if there isn't already one
Screenshot 2021-11-22 at 10 03 59
:

pjmonks:
Resolves #399


jamesrwelch (approved):
Nice one

Issue #350, opened 2021-10-06
Default dynamic profile view on data model/class/element level


jelenadreskai:
As a user, I want to be able to see s certain set of data model/class/element attributes by default (defined in one dynamic profile), so that information I need is readily available and I do not need to perform an additional search or click through different screens to see the information I need.

GBishop-PHSA:
@gammonpeter @jamesrwelch Could I get some feedback on whether this would be a good solution for Users to have a Custom Profile set as a default on Catalog Item Edit/View screens?

Proposed solution for setting a custom profile as the default selection in the Profile drop list
1. The UI will look for the default.profile property key on the current catalog item
	- The default.profile property key and namespace are determined by the Admin Configuration property entries of: 
		ui.property.namespace
		ui.property.default.profile.key
	- These properties will be loaded to the database through a new Grails bootstrap entry for dev/production environments
   a. If the default.profile property key is not found on the catalog item, the UI will search for this 
   property on the parent catalog item, and continue up the hierarchy until a property is found, or the top of
   the hierarchy of catalog items has been reached. (eg. Data Element > Data Class > Data Model)
2. The UI will set the Profile drop list value to: 
   a. Default if no default.profile property key is found in the hierarchy, or 
   b. Default if a default.profile property key is found but does not match an existing Custom Profile for the catalog item, or
   c. The Value found for the default.profile property key that matches an existing Custom Profile

jamesrwelch:
@GBishop-PHSA This seems like a really nice feature - I especially like that by setting the property in some of our custom imports we can control the way the model / class / element is presented.  I'm playing with the branch now and I can't seem to get it working - we've made some changes to the back end which may have inadvertently broken it?  Or maybe I've not followed the instructions above correctly.  Would you be able to take a look?

GBishop-PHSA:
@jamesrwelch I made a design change when we were testing the solution. Since the logic in 2c is "The Value found for the default.profile property value that matches an existing Custom Profile", it didn't seem necessary to have a configuraton to set the property key, just the namespace. 

So part 1, is now:
1. The UI will look for the default.profile property on the hierarchy of catalog items
The default.profile property namespace is determined by the Admin Configuration property entry of:
* ui.default.profile.namespace


eg. A Data Model that has a custom profile assigned called "Custom Profile A"
The Data Model would then have a property set:
   namespace               key              Value
   default.propfile        any_key        Custom Profile A

PR #437, opened 2022-01-20
gh-350 Set a custom profile to show in the profile on screen entry

GBishop-PHSA:
gh-350 Allow a default selection for the Profile droplist. The user can set a default.profile property at any level of the catalogue item hierarchy and the UI searches for these properties and sets default values in the profile droplist for Model, Class and Element if a Custom Profile name matches the property value. If no match is found, the existing 'default' selection is set in the droplist.


jamesrwelch:
@GBishop-PHSA This seems like a really nice feature - I especially like that by setting the property in some of our custom imports we can control the way the model / class / element is presented. I'm playing with the branch now and I can't seem to get it working - we've made some changes to the back end which may have inadvertently broken it? Or maybe I've not followed the instructions above correctly. Would you be able to take a look?

jamesrwelch (changes requested):
@GBishop-PHSA This works really nicely.  I've merged to the latest version of develop.  I wonder if you could just fix the few lint errors that are stopping a nice green tick?  Running
```
npm run eslint-report
```
will give you the report (eslint_report.xml) - don't worry about the warnings, but could you fix any 'error' messages on your new code?  I had a quick scan and it looks like they're all fairly straightforward to resolve.  Thanks!

GBishop-PHSA:
@jamesrwelch I've fixed the errors from the lint report

jamesrwelch (approved):
Lovely stuff - thank you very much!

PR #444, opened 2022-01-26
Remove validation constraints on codeset creation (Fixes #429)

### LINKED ISSUE #429: Remove requirement for a codeset to have at least one term
At the moment, there's a rules that states that a CodeSet must include at least one Term.  However, in a new project, we will want to record a Rule (a SNOMED Expression) for a CodeSet to calculate which terms are included (This calculation will happen externally to Mauro).  So for these CodeSets we don't want to include any terms explicitly.

jamesrwelch:
Remove the constraint that a codeset must have at least one term.  Also remove requirement for author and organisation to be completed


pjmonks (approved):
Tested without the author, organisation or term fields and it still creates the codeset as expected, so looks good.

Issue #442, opened 2022-01-26
Email notifications use catalog URL, which sends users to the missing page


jelenadreskai:
1. Reset User password
2. Email generated points to the variable "catalog URL"
3. Look at the email queue of failed messages
4.  Failed message shows a pointer to the catalog
5. When user clicks on the URL, user is taken to the undefined webpage
(Example: http://www.test.com/undefined)

![image](https://user-images.githubusercontent.com/81437684/151086967-493002ba-24d1-4365-b6d7-36271632e474.png)
![image](https://user-images.githubusercontent.com/81437684/151086902-8c14abb9-da18-4eff-a41c-6d68af4bfca1.png)

Environment: Mauro baseline 4.10.0/6.6.0

jelenadreskai:
It's a one line code change. Can be assigned to @GBishop-PHSA , she looked at it already.

jamesrwelch:
@jelenadreskai @GBishop-PHSA I'm not sure I understand - is this just a case of needing to change the catalogue URL variable through the admin interface?  Or is something else going on?

olliefreeman:
Yeah I agree, l this is actually a backend thing and is supposed to pull the info from
The api property site_url

jamesrwelch:
Ok, I see.  I guess the email log should really show the actual text sent, with the property values replaced?

olliefreeman:
@jamesrwelch  ??? the email log is the email exactly as its sent to the user stored into the DB. Therefore the property replacement has already been done.

jamesrwelch:
Ok, having looked at the API response and the UI rendering of the Url, I understand now.  Yes, @GBishop-PHSA please do fix this for us :)

GBishop-PHSA:
@jamesrwelch Ready for review. This change affects any text displayed in Mauro that contains a url (eg. Data Model HTML descriptions, email history).

PR #445, opened 2022-01-26
gh-442 Return original url from HTML link parser for http element types

GBishop-PHSA:
gh-442 Added processing for http element types when parsing text to html


jamesrwelch (approved):
Thanks - this works nicely for the issue reported.  There's still an issue for links to email addresses in text - I'll create a separate issue and assign it to you - would you be able to take a look while you understand that bit of the code?

GBishop-PHSA:
@jamesrwelch Yes, I can make a fix for email addresses as well in the new ticket.

Issue #446, opened 2022-01-26
Links to emails don't work correctly in email logs


jamesrwelch:
In the admin interface, in the email logs, email addresses appear as links, but when clicked go to 'undefined'. 
<img width="497" alt="Screenshot 2022-01-26 at 21 24 00" src="https://user-images.githubusercontent.com/1869668/151249463-40e0edd3-5f4b-4b16-93c4-9b25abf5c933.png">
Similar to #445 / #442, and so assigning to Gail, but I realise that's a bit presumptuous, so please feel free to unassign!

GBishop-PHSA:
@jamesrwelch  Ready for review.

PR #447, opened 2022-01-26
gh-446 Update HTML parser for mailto links

GBishop-PHSA:
gh-446 Update to link creator to return the mailto link for emails


jamesrwelch (approved):
Lovely stuff... thank you so much!

PR #450, opened 2022-01-28
Include version number of profiles when selecting to add/edit

### LINKED ISSUE #402: Indistinguishable dynamic profiles
**Describe the bug**
When adding a profile that has branches and versions to a data model multiple indistinguishable profiles are listed in the drop down

**To Reproduce**
1. Edit a profile with branching
2. Save
3. Try adding that profile to an existing model

**Expected behavior**
Only the current finalised profile is offered
**Screenshots**
image

pjmonks:
Resolves #402 by listing the version number next to the profile display name

image


jamesrwelch (approved):
Lovely :)

PR #457, opened 2022-02-02
Grey-out buttons after submission

shikhavani:
#456 # @olliefreeman @pjmonks

### DIFF
--- a/src/app/wizards/dataModel/data-model-main/data-model-main.component.ts
+++ b/src/app/wizards/dataModel/data-model-main/data-model-main.component.ts
@@ -51,6 +51,7 @@ export class DataModelMainComponent implements OnInit {
     private folders: FolderService,
     private title: Title) { }
 
+  disabledSave = false;
   ngOnInit() {
     this.title.setTitle('New Data Model');
 
@@ -90,6 +91,7 @@ export class DataModelMainComponent implements OnInit {
   };
 
   save() {
+    this.disabledSave =  true;
     const details = this.steps[0].compRef.instance as DataModelStep1Component;
     const types = this.steps[1].compRef.instance as DataModelStep2Component;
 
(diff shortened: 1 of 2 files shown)


src/app/wizards/dataModel/data-model-main/data-model-main.component.ts
```diff
@@ -114,12 +116,16 @@ export class DataModelMainComponent implements OnInit {
...
+          this.disabledSave =  false;
           this.messageHandler.showError('There was a problem saving the Data Model.', error);
           return EMPTY;
         }))
       .subscribe((response: DataModelDetailResponse) => {
         this.messageHandler.showSuccess('Data Model saved successfully.');
+        this.disabledSave =  false;
         this.stateHandler.Go('datamodel', { id: response.body.id }, { reload: true, location: true });
+      }, error => {
+        this.disabledSave =  false;
```
pjmonks, line 128: There is a shorter way of setting this flag back to `false` without having to duplicate it. There is an `rxjs` operator called `finalize()` which, if included in the `pipe()`, will always be called regardless of whether it was successful or not. Including something like the following would then you only need to set this flag once:

```ts
.addToFolder()
.pipe(
  catchError(error => {
    // Error case
  }),
  finalize(() => this.disabledSave = false)  // Always called at the end
)
.subscribe((response) => {
  // Success case
});
```

pjmonks (changes requested):
Generally the solution looks good, thank you. Please could you review my comments though to fix up a couple of things.

Also, there are linting issues found in your changes which our build picked up on. If you run this command you will get a HTML report, you'll only need to care about the errors found (there will be lots of warnings, please ignore those):

```bash
npm run eslint -- --format html -o eslint_report.html
```

@jamesrwelch You've mentioned that there may be more places where a submission button should be disabled whilst saving, perhaps the issue should be updated to list these places to clearly identify them? Then this PR could cover all those cases too.

shikhavani:
Ok. Sure. I will go through it and will update you

joe-crawford:
Thanks for your PR @shikhavani. Sorry for the delay, we'd missed this update. Would you be able to sign the Contributor Licence Agreement for this to be merged?

Link to the CLA: https://cla-assistant.io/MauroDataMapper/mdm-ui?pullRequest=457

cc @jamesrwelch

Issue #460, opened 2022-02-03
Data Elements of Data Types not displayed


joe-crawford:
# Description

When viewing the 'Data Elements' tab of a Data Type that is in use, no data elements appear.

# Steps to reproduce

1. Create a data model that has a data type which used by a data element, e.g. by importing an attached example model: 
[hr_model_20220203T135953.json.txt](https://github.com/MauroDataMapper/mdm-ui/files/7995135/hr_model_20220203T135953.json.txt)
2. Open the data type that is in use (Text type on Employees class in attached example)
3. Click on the Data Elements tab of the data type, no data elements appear.

# Expected behavior

The Data Elements tab of a Data Type should display the elements that use the type, accessible via the `{dataType}/dataElements` endpoint. 

# Environment

## Desktop 

Please complete the following information:
- OS: macOS
- Browser: Chrome 98

Tested on latest develop of mdm-ui and recent mdm-core.

jon-bowen:
dataModel was not being initialised in data-type.component. Added a subscription to initialise if exists.

jon-bowen:
Fixed whitespace issue.
I also added an npm script "eslint-nofix-quiet" so I can see what errors there are without being wiped out with the 811 warnings.

PR #468, opened 2022-02-15
gh-460: Data Elements of Data Types not displayed

jon-bowen:


pjmonks (changes requested):
The fix itself looks good and functions correctly 👍 

The only issue found during the build was a linting issue. Fortunately, if you run `npm run eslint` the error in question will automatically fix itself because it's so simple. If you could push up that change I'll be happy to approve.

jon-bowen:
Fixed whitespace issue.
I also added an npm script "eslint-nofix-quiet" so I can see what errors there are without being wiped out with the 811 warnings.

PR #469, opened 2022-02-16
MC-9802 Update tree control with endpoint updates

pjmonks:
* Folders returned for the tree root no longer auto-expand, so ensure that every container can be expanded on request
* Send modelCreatableOnly param for the tree view in the import screens

**Note:** this PR can be converted from draft once `MauroDataMapper/mdm-core` branch `feature/mc-9801` has been merged.

### DIFF
--- a/src/app/folders-tree/folders-tree.component.ts
+++ b/src/app/folders-tree/folders-tree.component.ts
@@ -283,19 +283,10 @@ export class FoldersTreeComponent implements OnChanges, OnDestroy {
     try {
       switch (node.domainType) {
         case CatalogueItemDomainType.Folder:
-          if (this.justShowFolders) {
-            const folderResponse = await this.resources.tree.get('folders', 'folders', node.id).toPromise();
-            return folderResponse.body;
-          } else {
-            return node.children;
-          }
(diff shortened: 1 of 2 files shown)


olliefreeman (approved):
Develop core branch is merged. This is good to go from my side

jamesrwelch:
Folder contents are no-longer appearing in alphabetical order.  Is an extra parameter to the endpoint required?  @olliefreeman @pjmonks
Otherwise, seems to work just fine.  I wonder whether a future improvement might be to retrieve and cache some of the folders in the background, so that expanding is quicker.  But overall a nice speed improvement.
Screenshot 2022-02-17 at 14 14 19

olliefreeman:
theres no optoins to change it its hardcoded
folders are always alphabetical
models should be as well
however DM content (DCs) "should" be returned as alphabetical childless then alphabetical parents

however ... i saw this exact problem the other day...im wondering if something i changed somehow caused the ordering ("sort") call to not be made, prob best to raise a bug on mdm-core and say that folders do not appear to being sorted properly in the tree. Then i'll fix it asap

PR #473, opened 2022-02-23
Feature/gh 466

jon-bowen:
Url routing parameter dataClassId was not being extracted from the fnode, and the dataModelId was not being passed into the url matching anyway.
There was another bug which has also been fixed by this:
Expand a folder (folder 1) and another folder (folder 2)
Click on a data class in folder 2 so that it's shown in the right hand pane; then right click on a data class in folder 1 and click View => Open in new window. The result will be "not found".


pjmonks (approved):
This seems to fix the issue. Only minor thing I've noticed is that line 587 seems to be indented twice compared to the rest of the code - might be worth checking your tab setup in your editor, it should be 2 spaces for 1 tab. That's not going to prevent me approving this though 👍

Issue #52, opened 2021-04-22
Merging Only works for DM


gammonpeter:
Trying to merge anything but a DM fails

PR #485, opened 2022-03-03
#483 Use correct endpoint for moving Folders and Versioned Folders

### LINKED ISSUE #483: Changing a Folder's parent is using the wrong endpoint which is stopping finalised VFs from being moved
When changing a folder's parent the UI should NOT be using a standard PUT/update with the body
```json
{"id":"23fc4906-c1a9-43fc-8b8f-7414136ff426","parentFolder":"fc6a86e3-2497-4f81-a790-2fc9de387d71"}
```

it MUST use the changeFolder endpoint `http://localhost/api/folders/{{folder_id}}/folder/{{destination_folder_id}}`
if the above endpoint is not used then VersionedFolders cannot be moved around after they are finalised.

There is a corresponding issue to update this endpoint for "root" moves where the `destination_folder_id` would be null see MauroDataMapper/mdm-core#235.

pjmonks:
Fixes #483 so that Versioned Folders can now be moved after they have been finalised.

Depends on PR MauroDataMapper/mdm-resources#52


jamesrwelch (approved):
Nice one 👍

PR #486, opened 2022-03-03
#464 Use content editor for description fields in "New..." wizards

### LINKED ISSUE #464: Description field editor should be HTML editor when creating things
When creating a data model, data class, data element, etc there is the opportunity to enter a description - this text field editor should us the Jodit editor to allow HTML / Markdown to be added
Description Editor

pjmonks:
Resolves #464 

image


jamesrwelch (changes requested):
Minor change - I thin it needs the word 'description' somewhere so we can see which field is being edited.

pjmonks:
I've tried to add labels above the description content editors now, which look like this:

image

They don't look consistent with the other Angular Material fields because they are not wrapped in a `mat-form-field` component - they can't be, because these custom editors have not been configured to be allowed as child components. I've tried to replicate the style of these labels as best I can for now instead.

jamesrwelch (approved):
Yes, this is great, thanks :)

PR #487, opened 2022-03-04
Change requests to get administrator status of user

### LINKED ISSUE #484: Remove the additional call during the login process to allow the login window to disappear
Currently when logging in the UI makes an additional call to `api/session/isApplicationAdministration` as soon as it receives confirmation from the API that the user is logged in. It will not close the login popup window until the response from this call is received.

The issue is that an improvement to the API to speed up the log in process itself is now returning the first call but leaving the API in a "locked" state for the user's security policy. Any subsequent calls to the API will stall until the lock has been removed via the async operation building the security policy.

This means the call to this additional endpoint is stalling for the time the policy is locked, thereby not progressing the user any further into the application.

If the UI did not make this call immediately, but rather closed the popup window and then progressed and made the call while it loaded the tree then it could allow users to possibly perform other actions.

No additional actions or endpoints are required to handle the "lock" its a built in feature to the API to just hold any calls while a policy is being updated.

pjmonks:
Resolves #484 

* Removed `api/session/isApplicationAdministration` endpoint request within login operation to speed up login. Backend will now do security work in the background and not block. This closes the Login dialog faster and speeds up the UI.
* Updated all call sites to use correct way of request is administrator state
* Cache `isAdmin` flag so subsequent requests will not send unnecessary network requests

You can test this on the MauroDataMapper/mdm-core branch `feature/mc-9803` to see the full speed benefit, though this change is not dependent on that - you can run the latest snapshot of `mdm-core` and the UI changes will still work. The key part is that all previous functionality that must check whether the user is an admin or not works exactly as before.

### DIFF
--- a/src/app/admin/app-container/app-container.component.html
+++ b/src/app/admin/app-container/app-container.component.html
@@ -48,7 +48,7 @@ SPDX-License-Identifier: Apache-2.0
                 </li>
             </ul>
 
-            <ul class="menu-list" *ngIf="isAdmin">
+            <ul class="menu-list" *ngIf="isAdministrator">
                 <p class="marginless pl-2 text-muted">Admin settings</p>
                 <li>
                     <a role="menuitem" uiSref="appContainer.adminArea.home" uiSrefActive="active">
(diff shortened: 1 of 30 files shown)


olliefreeman:
MauroDataMapper/mdm-core#283 contains the API changes

olliefreeman:
MauroDataMapper/mdm-core#283 is in develop

jamesrwelch (approved):
This all works very nicely and the UI feels a lot slicker now 👍

PR #488, opened 2022-03-09
#407 Choose terminology for codeset now returns full ordered list

### LINKED ISSUE #407: When editing a codeset, and choosing a terminology, list is not properly paginated or ordered
When editing a codeset, and choosing a terminology, list is not properly paginated or ordered.  This means the search doesn't work either

Screenshot 2021-12-15 at 09 22 55

pjmonks:
Resolves #407 

The call to fetch the available terminologies was simply missing query parameters to return all (no pagination) and set correct sort order of results.


jamesrwelch (approved):
Great, thanks.  I think additionally you'd want to see the version no. where multiple terminologies exist with the same version, but that can be another ticket.

Issue #493, opened 2022-03-14
Indistinguishable dynamic profiles when viewing


joe-crawford:
# Description

Similar to #402 (when editing), when viewing a dynamic profile with multiple versions on a data model, the versions are indistinguishable.

# Steps to reproduce

1. Using the UI, create a dynamic profile specification model and finalise (to create version 1.0.0).
2. Create a new version of the dynamic profile specification model from 1. and finalise again (to create version 2.0.0).
3. Create a data model and apply one of the dynamic profile versions.
4. When viewing the data model, in the profile dropdown, the two dynamic profile versions are indistinguishable.

# Expected behavior

If multiple profile versions are shown when viewing a datamodel, the version number should be included (as it is when adding a profile after PR #450).

# Environment

## Desktop 

Please complete the following information:
- OS: macOS
- Browser: Chrome 99
- Mauro: latest develop of core, ui and resources

# Additional context

N.B. viewing multiple profile versions using develop of mdm-core may be affected by maurodatamapper/mdm-core#292 if there is a non-finalised profile specification model version.

joe-crawford:
Also change the View/Edit and Add New profile dropdowns to only show the latest (finalised) versions of profiles.

Issue #56, opened 2021-04-22
Empty string MDM_UI_THEME_NAME env variable breaks the UI theme loading


olliefreeman:
**Describe the bug**
If `MDM_UI_THEME_NAME` is set to an empty string then the UI cannot load any theme resulting in a 404 on `assets/themes//logo.png`

**To Reproduce**
Set `MDM_UI_THEME_NAME=''` then run or build and run the ui

**Expected behavior**
If the env variable is blank or null or not present then we should use the default theme, we should also probably handle an unknown theme to default to the "default" theme incase someone puts junk in.

**Screenshots**
![Screenshot-2021-04-22-at-12 20 47](https://user-images.githubusercontent.com/2587424/115706676-ceb62500-a365-11eb-9aec-0e8289f358c0.png)

PR #494, opened 2022-03-17
Show version numbers in view/edit profiles dropdown

joe-crawford:
Add version number to profile view/edit dropdown and allow backend to filter versions of profiles displayed (resolves gh-493)

### DIFF
--- a/src/app/modals/add-profile-modal/add-profile-modal.component.ts
+++ b/src/app/modals/add-profile-modal/add-profile-modal.component.ts
@@ -44,7 +44,7 @@ export class AddProfileModalComponent implements OnInit {
         profiles.body.forEach((profile) => {
           const prof: any = [];
           prof['display'] = `${profile.displayName} (${profile.version})`;
-          prof['value'] = `${profile.namespace}/${profile.name}`;
+          prof['value'] = `${profile.namespace}/${profile.name}/${profile.version}`;
           this.allUnusedProfiles.push(prof);
           this.showProfileSelector = true;
         });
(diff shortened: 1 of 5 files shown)


joe-crawford:
Depends on https://github.com/MauroDataMapper/mdm-resources/pull/56.

Also depends on https://github.com/MauroDataMapper/mdm-core/pull/293 (merged to develop).

jamesrwelch (changes requested):
One thing that seems to have stopped working with this commit... is that when you add a new profile, it doesn't properly route the next page to display the new profile properly... Are you able to reproduce this and work out this issue?

joe-crawford:
> One thing that seems to have stopped working with this commit... is that when you add a new profile, it doesn't properly route the next page to display the new profile properly... Are you able to reproduce this and work out this issue?

Yes I had noticed similar issues, although I think they were already occurring on develop branch (raised as gh-503). Looking again I have realised it is related to how the new configurable default profile is set. The above commit resolves it for me however as the issue was sometimes intermittent, so could you check if this resolves it for you?

joe-crawford:
Commit 864cba2 contains a change to gh-350 Default dynamic profile view configuration:

The API property to set is still `ui.default.profile.namespace`; it must be publicly readable.

As the profile display names may not be unique, the default profiles are now set by:

| namespace | key | value | example value |
|-|-|-|-|
| value of `ui.default.profile.namespace` | `namespace` | profile provider `namespace` | `uk.ac.ox.softeng.maurodatamapper.profile.provider` |
| value of `ui.default.profile.namespace` | `name` | profile provider `name` | `A%20Dynamic%20Profile` |
| value of `ui.default.profile.namespace` | `version` | profile provider `version` | `1.0.0` |

jamesrwelch (approved):
Amazing work, well done!

PR #496, opened 2022-03-22
#465 Add external link option to Jodit editor

### LINKED ISSUE #465: HTML Editor doesn't allow external links to be added
The 'Add link' button on the HTML editor allows you to create a link to another Mauro object, but no-longer allows you to create a link to an external URL.  This should be re-enabled

pjmonks:
Add both toolbar options to insert internal (catalogue items) and external links.

Resolves #465 

image


jamesrwelch (approved):
Yes, this works.  I'm not sure about the 'attachment' icon, but can't immediately think of anything better...

PR #497, opened 2022-03-22
#476 Updates to Create Term dialog

### LINKED ISSUE #476: Multiple updates to 'Create term' window
When creating a new term, the following window appears:
Screenshot 2022-02-24 at 11 48 40

The following changes are required:
- The window should be larger by default
- The 'description' field should be an html editor
- The 'definition' field should be optional - if no definition is supplied then the API call should set the definition to be the same as the code.
- After creating the term, the UI returns to the terminology page - this should follow the same pattern as other 'create' methods and send you to the description tab of the newly-created term.

pjmonks:
* Adjust window size to be bigger
* Description field now use a HTML/Markdown editor
* Definition field is now optional
* Once created, you are redirected to the new term created

Resolves #476 

image

### DIFF
--- a/src/app/model/defaultProfileModel.ts
+++ b/src/app/model/defaultProfileModel.ts
@@ -82,6 +82,7 @@ export  class DefaultProfileControls {
 
     const term = [
       'code',
+      'definition',
       'description',
       'aliases',
       'classifications',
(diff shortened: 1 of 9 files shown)


jamesrwelch (approved):
Yep, works nicely, thank you

PR #498, opened 2022-03-22
#481 Add term code to default profile to allow edits

### LINKED ISSUE #481: Edit the code of a term
At the moment it's not possible to edit the code of an existing term.  You can edit the definition by using the '...' drop-down in the top right (that actually says 'edit title').  But not the code.

It would be nice to edit both code and definition together in the same window.  The standard 'edit' window may be sufficient.

pjmonks:
By adding the term code to the default profile, selecting "Edit" allows the user to make changes to it.

Resolves #481 

image


jamesrwelch (approved):
Brilliant, works nicely, thanks!

PR #508, opened 2022-04-11
gh-499 Add user preference for content editing format

### LINKED ISSUE #499: User preference for HTML / Markdown
It would be nice if a user could set a preference for whether the HTML editor opens up for new / blank text in HTML or Markdown mode.  This could be done in the user preferences section of the UI and applied whenever the text editor is opened with empty text.

pjmonks:
Resolves #499 

* Add type definitions to UserSettingsHandlerService. Add missing properties that were used and refactor to not mix promises
* Update user preferences page
* Update content editor to respect user preferences. If no preference is set, or is "auto", revert back to original behaviour of guessing appropriate format

User has choice between forcing content to be Markdown or HTML. If they pick "Automatic" (or don't make a choice), behaviour reverts to original, where format is best-guessed from existing text values.

image

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -20,8 +20,8 @@ import { OverlayContainer } from '@angular/cdk/overlay';
 import { Component, HostListener, OnDestroy, OnInit } from '@angular/core';
 import { UserIdleService } from 'angular-user-idle';
 import { ToastrService } from 'ngx-toastr';
-import { Subject } from 'rxjs';
-import { takeUntil } from 'rxjs/operators';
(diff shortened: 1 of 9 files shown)


jamesrwelch (commented):
I've had a play, and made some changes:
- If a field already has content, it determines whether the content is html or markdown and opens the editor in that mode
- If a field is empty, it uses the user preferences to determine which mode to open the editor in
- Users can always change mode
- The 'auto' option has been dropped, and markdown is the default for the preference.

jamesrwelch (approved):
Lovely work.  I made changes based on me having a play and getting my requirements sorted, but the whole thing works like a charm

Issue #501, opened 2022-03-26
"About" button flexibility


oughnic:
# Description

Make the "About" button destination parameterised so that it can point to a page describing the use of Mauro in a specific context rather than pointing to a detailed page about Mauro.

# Additional context

As Mauro becomes a platform for specific uses it looses the need to be explaining what Mauro is, and becomes a content management system for data models, metadata, master data, reference data and rules.  Mauro's distinctive capability is to map between models and flesh out the context.  A user on a site is not going to want to learn about Mauro, in the same way as they don't want to learn about Wordpress to use a Wordpress site.

# See also

"Learn more" button flexibility

jelenadreskai:
@jamesrwelch were there any plans to get this picked up by someone. If not, @GBishop-PHSA could implement this and contribute, it would be configurable, the same way the main page is configurable.  We had the exactly the same requirement on our side, and currently we overwrite Mauro's code. Ideally, we do not want to have **any** _custom_ code in future, just pick up new version of Mauro, configure our PHSA specific properties, and it would work out of the box. Does it make sense?

jamesrwelch:
Hi Jelena, 
If you and @GBishop-PHSA would like to pick this up, that would be fantastic, thank you.  We have a small package of UI updates that we're hoping to get built over the next couple of months; this was on them, but if you were able to take this one, we could concentrate on some others (e.g. searching, dashboards) that are also on the list.

PR #517, opened 2022-04-28
GH-501 Update About page to have configurable content

GBishop-PHSA:
gh-501 Removed html from about page and replaced with calls to new api system properties:
content.about.application.name - the name of the application that appears as the page header as a string value
content.about.content - the body of the about page as an HTML string
content.about.contact.details - the last line on the about page containing contact info as an HTML string


pjmonks (approved):
I'm happy with this. It functions in a similar way to modifying the homepage and the previous content acts as the default values if these new API properties are not present.

Pass from me ✔️

PR #533, opened 2022-05-16
gh-522 Update footer layout

### LINKED ISSUE #522: Create a consistent page footer
# Description

The footer at the bottom of every page in Mauro Data Mapper has been identified as needing some UX improvements:

1. It is hard to visually tell it is a footer - the main header/navigation area has a distinctive dark blue colour, but the footer is white with only a minimal border line to distinguish between the page content and the footer.
2. The alignment of the components within the footer are also not consistent when compared with the header.
3. A new Mauro Data Explorer UI is being created with a consistent UX/design, so the Mauro Data Mapper should be updated to reflect the same design elements at least.

The new footer should:

1. Not shift footer links to the far right, instead align them to the left
2. Allow some level of customisation e.g. if a copyright notice needs to change. Allow this via Mauro API properties.
3. Maintain the same colour scheme as the header (or whichever theme has been selected for the application).

# Additional context

Existing footer layout:

image

Mockup of intended new footer:

image

pjmonks:
Resolves #522 

* Design now replicates MDE
* Theming of CSS remains intact - can switch theme to "NHS Digital" too
* Footer links are passed down from the `app.component` now
* Added tests
* Updating package config for running tests and PR checks

image


jamesrwelch (approved):
I've had a play with this and it works nicely.  I've opened a separate issue (#548) to review and update the default contents of the footer, but I'm happy with the new footer component and its reuse

Issue #545, opened 2022-05-20
Add a button to remove Terms from a CodeSet


jamesrwelch:
At the moment there is no button to allow you to remove terms from a codeset.  Ideally this would be a little 'bin' symbol on the right-hand side of the term list.  Options to delete in bulk would be a nice extra.

Not sure if there's already a ticket for this, but I couldn't find one

PR #549, opened 2022-05-27
Set the access when configuring the code set table

jamesrwelch:
Fix for #545

### DIFF
--- a/src/app/code-set/code-set/code-set.component.html
+++ b/src/app/code-set/code-set/code-set.component.html
@@ -16,7 +16,7 @@ limitations under the License.
 
 SPDX-License-Identifier: Apache-2.0
 -->
-<div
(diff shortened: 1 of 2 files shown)


jamesrwelch:
The test failure highlights some inconsistencies with the way pages and tabs are generated - I won't try and fix them all here but will try and get this particular page working well enough to copy for other components

jamesrwelch:
Sorry, messed this one up a bit - it's now ready for review.
Probably need a chat about unifying components at some point!

pjmonks (approved):
This looks good and seems to function as required.

Consolidation of the components would be a good idea long term, but also a big project - there are certainly a lot of components that carry out a similar purpose.

This fix is approved ✔️

PR #552, opened 2022-05-27
Feature/gh 491

### LINKED ISSUE #491: Improve html description rendering in list view
In the list view, html descriptions are often not clearly rendered due to our naive way of only showing a subset of the content.  

Here's a SO post which contains a similar problem, and a potential solution:

https://stackoverflow.com/questions/12307666/how-to-create-a-show-more-button-and-specify-how-many-lines-of-text-can-be-ini

Solution demo:
https://jsfiddle.net/Wpn94/

Maybe we could build the same thing into our tables somehow.

jamesrwelch:
Fixes #491 - See ticket for screenshots of new functionality.


src/app/shared/more-description/more-description.component.ts
```diff
@@ -16,52 +16,42 @@ limitations under the License.
...
-    const desc = this.markdownParser.parse(this.description, 'text');
-    if (desc && desc.length > this.maxLength) {
-      let subStr = desc.substring(0, this.maxLength);
-      const lastIndexOf = subStr.lastIndexOf(' ');
-      subStr = subStr.substring(0, lastIndexOf);
-      return `${subStr}...`;
-    } else {
-      return desc;
-    }
+  checkOverflow (element): boolean {
```
pjmonks, line 53: This function is only used internally to the component, so can be marked as `private`.

src/app/shared/more-description/more-description.component.ts
```diff
@@ -16,52 +16,42 @@ limitations under the License.
...
+
   constructor(
-    userSettingsHandler: UserSettingsHandlerService,
-    private markdownParser: MarkdownParserService
+    userSettingsHandler: UserSettingsHandlerService
   ) {
     this.showMore = userSettingsHandler.get('expandMoreDescription');
   }
 
   ngOnInit() {
```
pjmonks, line 42: If `ngOnInit()` is empty now, it can just be removed along with the `OnInit` interface.

pjmonks (changes requested):
Generally it looks and functions good. If you're in the area of updating the existing `mdm-more-description` component, now would be a good time to tidy it up.

Also, please could you confirm if this change is only meant for the email list or for other table components too?

jamesrwelch:
It's a replacement for the old component, so it should be in use everywhere.  The email list was the only place a change was required (removing the length parameter which is no longer a feature).

I'll tidy as you suggest.

src/app/shared/more-description/more-description.component.ts
```diff
@@ -16,52 +16,42 @@ limitations under the License.
...
 
 @Component({
   selector: 'mdm-more-description',
   templateUrl: './more-description.component.html',
   styleUrls: ['./more-description.component.sass']
 })
-export class MoreDescriptionComponent implements OnInit {
+export class MoreDescriptionComponent implements OnInit, AfterViewInit {
   @Input() description: string;
   @Input() length: any;
```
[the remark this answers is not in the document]
jamesrwelch: Yes - good call.  Highlighted another couple of places where it was used, so fixed those too

src/app/shared/more-description/more-description.component.ts
jamesrwelch: Consider it privated.

src/app/shared/more-description/more-description.component.ts
jamesrwelch: I've now made this change

jamesrwelch:
@pjmonks I think this is ready for review again now 👍

pjmonks (approved):
All looks great to me, thanks for the updates 👍 

Approved ✔️

PR #560, opened 2022-06-08
Feature/gh 527 - Pagination on search listing

### LINKED ISSUE #527: Add pagination to Catalogue Search
# Description

Once #525 has been completed, pagination controls should be added to the search results list to be able to step through multiple pages of search results. This will also reduce search result response sizes by limiting the data returned by Mauro.

# Additional context

* The Mauro Data Explorer has a service and component to render numbered page buttons which can be reused.
* The same `max` and `offset` parameters should be sent in endpoint requests to Mauro to control the pagination.

aaronforshaw:
Add paginator control to catalogue search listing page.

Closes #527

### DIFF
--- a/src/app/catalogue-search/catalogue-search-listing/catalogue-search-listing.component.ts
+++ b/src/app/catalogue-search/catalogue-search-listing/catalogue-search-listing.component.ts
@@ -16,7 +16,7 @@ limitations under the License.
 
 SPDX-License-Identifier: Apache-2.0
 */
-import { Component, OnInit } from '@angular/core';
+import { Component, OnInit, ViewChild } from '@angular/core';
 import { MessageHandlerService, StateHandlerService } from '@mdm/services';
 import { UIRouterGlobals } from '@uirouter/core';
 import { EMPTY } from 'rxjs';
@@ -27,6 +27,7 @@ import {
   CatalogueSearchResultSet,
   mapStateParamsToSearchParameters
 } from '../catalogue-search.types';
+import { MdmPaginatorComponent } from '@mdm/shared/mdm-paginator/mdm-paginator';
 
 export type SearchListingStatus = 'init' | 'loading' | 'ready' | 'error';
 
@@ -36,6 +37,7 @@ export type SearchListingStatus = 'init' | 'loading' | 'ready' | 'error';
   styleUrls: ['./catalogue-search-listing.component.scss']
 })
 export class CatalogueSearchListingComponent implements OnInit {
+  @ViewChild(MdmPaginatorComponent, { static: true }) paginator: MdmPaginatorComponent;
   status: SearchListingStatus = 'init';
   parameters: CatalogueSearchParameters = {};
   searchTerms?: string;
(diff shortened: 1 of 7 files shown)


src/app/catalogue-search/catalogue-search-listing/catalogue-search-listing.component.ts
```diff
@@ -68,6 +72,12 @@ export class CatalogueSearchListingComponent implements OnInit {
     });
   }
 
+  onPageChange() {
+    this.parameters.page = this.paginator.pageIndex;
+    this.parameters.pageSize = this.paginator.pageSize;
+    this.performSearch();
```
pjmonks, line 78: This works fine, but it wasn't what I was expecting. I apologise if my intentions were not clear in the issue description.

I've worked on a lot of e-commerce websites which generally have this pattern of displaying search results/listings. What I learned from that is the URL to the page should explain exactly what the search should do, which is why you usually see very long query strings on search pages. The reasoning behind this is because of use cases like:

1. A user might bookmark that page to come back to later
2. Someone might copy the link to that search results page and send to someone else in a message/email
3. SEO reasons (though probably not applicable to MDM)
4. etc

So if there was a URL `https://my.mauro.org/#/search/listing?search=test&page=2&size=20` then this should always return the 2nd page of search results for "test", showing the next 20 results. This is how I implemented the MDE search page, but this version in `mdm-ui` does not do that because each time you change pages the URL is not updated. At the moment, if I passed in a query param of `page=2`, I would still initially see the first page of results.

This is easily fixed though. If instead of re-rendering the search results list, you instead could use the UI router to navigate to the same page route but with the updated query params - see the `SearchListingComponent` in `mdm-explorer` for what I mean. This should then render a new `CatalogueSearchListingComponent` and trigger the `ngOnInit()` to start the next search call.

Because `mdm-ui` uses UI Router though instead of Angular router, I can't remember if just transitioning the page route is enough, you might need to also pass a `{ reload: true }` option to the router to force a reload - something you'll have to test.

src/app/catalogue-search/catalogue-search-listing/catalogue-search-listing.component.ts
```diff
@@ -52,6 +54,8 @@ export class CatalogueSearchListingComponent implements OnInit {
     this.parameters = mapStateParamsToSearchParameters(
       this.routerGlobals.params
     );
+    this.parameters.page = this.paginator?.pageIndex ?? 0;
```
pjmonks, line 57: If redirecting to the page itself when pagination changes as I explained in another comment, then I don't think accessing the `paginator` component would be required, because the query parameters will dictate the page and page size. Suitable defaults will still need to be used in case those query params are missing/undefined though.

pjmonks (changes requested):
Could you look over the comments I've left please? I hope what I've suggested is OK.

src/app/catalogue-search/catalogue-search-listing/catalogue-search-listing.component.ts
```diff
@@ -68,6 +72,12 @@ export class CatalogueSearchListingComponent implements OnInit {
     });
   }
 
+  onPageChange() {
```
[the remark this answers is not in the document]
aaronforshaw: Done

src/app/catalogue-search/catalogue-search-listing/catalogue-search-listing.component.ts
aaronforshaw: Thanks for the info. As well as changing the route to accept an additional `pageSize` query parameter, I also had to change `MdmPaginatorComponent` to:
(a) accept a `[pageIndex]` input
(b) properly deal with the `[pageSize]` input i.e. rather than always setting pageSize to 20, use the pageSize from the input, or failing that from user settings, else default to 20.

pjmonks (approved):
Fantastic, thanks for the changes 👍 ✔️

PR #562, opened 2022-06-17
MC-9841 Update UI to support certain actions that can run async in Mauro

pjmonks:
# Details

* Update fork/branch/version of models to allow asynchronous creation
* Update versioned folders to support async fork/branch/version
* Updates to Import Models page
* Add async option to Export Models page
* Update model detail pages to support exports. Use a dialog to select export options now.
* Prevent export models from mixing domain types in model selection
* Use mdm-download-links component for Export Models page
* Update infrastructure code to get correct MDM resource easier
* Refactor other related code for type correctness

All async actions will display an info toast, if successful, stating the background job has started.

# Screenshots

Branching/versioning/forking:

image

Export model (from model detail - new dialog):

image

Export models (multiple):

image

Import models - parameters pulled directly from importer plugin:

image

### DIFF
--- a/src/app/dataModel/data-model-detail.component.html
+++ b/src/app/dataModel/data-model-detail.component.html
@@ -209,23 +209,14 @@ SPDX-License-Identifier: Apache-2.0
             color="primary"
             type="button"
             *ngIf="isLoggedIn"
-            class="paddingless dropdown-toggle"
+            class="paddingless"
             [disabled]="processing || deleteInProgress"
-            [matMenuTriggerFor]="exporterList"
             matTooltip="Export as JSON, XML,..."
             aria-label="Export as JSON, XML,..."
+            (click)="exportModel()"
           >
             <span class="fas fa-download"></span>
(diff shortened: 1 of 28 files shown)


jamesrwelch (dismissed):
This seems to work nicely.  One minor request - after an export the download link (in the admin 'jobs' section) is not very user-friendly - could we have a 'Download file' button or something?

pjmonks:
Updated as best I can by rendering the async job messages as HTML:

image

This does produce a useable hyperlink which triggers a download. But we can't have a dedicated button for this because the async job details returned from Mauro do not provide any further context - such as the domain export ID that it is linked to, or additional context extracted from the message.

jamesrwelch:
Minor question / comment - when I download a file using the new link, it opens a new blank browser tab at the same time as downloading.  Is that to do with my browser?
Otherwise I'm happy with this.

jamesrwelch (approved):

PR #563, opened 2022-06-17
MC-9840 Display async jobs for a user

pjmonks:
# Details

* New page added to user profile area
* Supports filtering, sorting and pagination
* View and cancel async jobs

# Screenshots

List async jobs available:

image

Details on a single job:

image


jamesrwelch (approved):
Really nice - I like it!  Thanks!

PR #564, opened 2022-06-17
MC-9889 Display domain models exported asynchronously

pjmonks:
# Details

* New page added to user profile area
* Supports filtering, sorting and pagination
* View, download and delete exports

# Screenshots

List available exported models:

image

Details of an export:

image


jamesrwelch (approved):
Works very nicely, good stuff!

PR #566, opened 2022-06-21
Feature/gh 529 Partial implementation of search filtering

aaronforshaw:
Add search filtering by domainTypes, labelOnly, exactMatch, lastCreatedBefore, lastCreatedAfter, createdBefore, createdAfter.

Still to do: styling, filter by a specific catalogue item, reset search (i.e. clear all filters).


src/app/catalogue-search/catalogue-search-listing/catalogue-search-listing.component.ts
```diff
@@ -113,6 +117,16 @@ export class CatalogueSearchListingComponent implements OnInit {
     this.updateSearch();
   }
 
+  onFilterChanged(event: SearchFilterChange) {
+    console.log(event);
```
pjmonks, line 121: The `console.log()` line can be removed.

src/app/catalogue-search/search-filters/search-filters.component.scss
```diff
@@ -0,0 +1,65 @@
...
+SPDX-License-Identifier: Apache-2.0
+*/
+ul {
+  list-style-type: none;
+  margin-top: 4px;
+  padding-left: 0;
+  padding-right: 0;
+}
+
+/*$search-filters-heading-background-color: $color-mauro-dark-green;
```
pjmonks, line 26: I'll assume these commented out styles will get tidied up later...

pjmonks (approved):
Only minor comments and UI suggestions, but generally this looks great and is definitely heading in the right direction ✔️ 

If my comments could be addressed in the next iteration, as discussed, then I'll approve this now.

aaronforshaw:
@pjmonks should the "Clear All" button reset the Label Only filter to its default value of true, or really clear it to false?

pjmonks:
@aaronforshaw Good point, "clear" would suggest the "label only" checkbox goes back to an empty, unticked state. Perhaps the button should be called "Reset all" instead, so it's clear that all filters go back to their default state.

PR #569, opened 2022-06-23
Feature/gh 529 ux - Improvements to catalogue item search filters

aaronforshaw:
Improve catalogue item search filters as follows:
- Tidy up styles and alignments
- Add a 'Reset All' button
- Default labelOnly to true
- Change domainTypes filter to a mat-select
- Add a 'clear' button to the date fields

Known bug: if you click the 'Reset All' button twice in succession, on the second click the router seems to hang

Still to do: Add tree control to filter by a specific catalogue item

### DIFF
--- a/src/app/catalogue-search/catalogue-search-listing/catalogue-search-listing.component.html
+++ b/src/app/catalogue-search/catalogue-search-listing/catalogue-search-listing.component.html
@@ -47,7 +47,10 @@ SPDX-License-Identifier: Apache-2.0
   </div>
   <div *ngIf="status === 'ready'">
     <div class="row">
-      <div class="col">
+      <div class="col-md-3">
+
+      </div>
+      <div class="col-md-9">
         <div class="mdm-search-listing__sort-row">
           <div class="vertically-aligned">
             {{ resultSet?.totalResults }} results found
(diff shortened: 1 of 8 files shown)


pjmonks:
I've noticed an unusual bug when I was just clicking around the filters:

# Steps to reproduce

1. Go to a search results page e.g. http://localhost:4200/#/search/listing?search=episode&labelOnly=true
2. Click on the "clear" button inside a date filter - at the moment it shouldn't have a value:

image

3. Clicking the button should trigger the search again, with the same results
4. Now click the "clear" button again

# Expected

Another search should be fired off, again with the same results

# Actual

image

The loading spinner remains in place so nothing is loaded, the page is stuck. The odd thing is that there are no errors listed in the browser network calls or console, so no obvious exception is happening.

If you'd like to separate this out into it's own issue to look at, please do.

pjmonks:
@aaronforshaw Sorry, I should have read your PR description before explaining a bug you already knew about 🤦‍♂️ 

Please feel free to record the bug as a separate GH issue if you want though, as long as it gets looked at it doesn't look like it should block approval of this PR.

pjmonks (approved):
This all looks great! 🎨 

The clear/reset bug seems strange, but can be worked on separately.

Approved ✔️

PR #575, opened 2022-06-29
gh-570 Fix search reset bug which sticks on loading screen

### LINKED ISSUE #570: Catalogue item search can hang
# Description

The catalogue item search page can be provoked to hang with a loading spinner.

See description in https://github.com/MauroDataMapper/mdm-ui/pull/569

pjmonks:
Fixes #570 

Due to the way the UI Router works when it decides no transition between routes is required, the loading screen was accidentally kept on preventing the user from doing anything else. This PR resolves that issue.


aaronforshaw (approved):
Excellent 👍

Issue #70, opened 2021-04-29
Unable to Export Reference Data Model


gammonpeter:
**Describe the bug**
When trying to export Reference Data Models an error with renderer prevents it from completing

PR #586, opened 2022-07-01
gh-530 Update search within catalogue item detail views

### LINKED ISSUE #530: Update search within individual catalogue items
# Description

Individual catalogue items can also display search results, restricting results to that top-level catalogue item. This is currently available for:

* Data Model
* Data Class
* Classification
* Folder
* Versioned Folder

This uses the `<mdm-advanced-search-bar>` component, which is intended to be phased out in #531. Instead, replace catalogue item search functionality with the new search components created for the Catalogue Search page.

# Additional context

* The existing layout should remain - clicking on a "magnifier" icon will expand a Search panel, then list results under it, like so:

image

* Create a new container component to hold/render all common `CatalogueSearchModule` components, to render:
    * A search bar
    * Limited advanced filters:
        * Domain types to return - limit depending on top-level domain type
        * Label only - yes/no
        * Exact match - yes/no
        * Last updated - date range
        * Created - date range     
* Pagination and sort controls - allow display of up to 10 results per page

pjmonks:
Resolves #530

**Note:** depends on PR MauroDataMapper/mdm-resources#70

image

* Remove original, legacy search controls from detail views
* Add new "Search" tab
* Create new component `mdm-catalogue-item-search` to contain search results
* Update CatalogueSearchService and add "refine search" link. Search service now uses context to decide which root item search endpoint to use
* Supported detail views are Data Model, Folder and Versioned Folder

### DIFF
--- a/src/app/import-models/import-models.component.html
+++ b/src/app/import-models/import-models.component.html
@@ -55,7 +55,7 @@ SPDX-License-Identifier: Apache-2.0
           >
             <mat-option *ngFor="let item of importers" [value]="item">{{
               item.displayName
-            }}</mat-option>
+              }}</mat-option>
           </mat-select>
         </mat-form-field>
       </div>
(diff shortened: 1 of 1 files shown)


jamesrwelch (approved):
This works nicely, and looks good.  I have two minor comments:
1. The description field might be better to use the new 'show more / less' component, because long descriptions make the page rather long
2. The `mat-progress-spinner` component never works nicely on Safari - it jumps about all over the place.  Not just here, but all over the site.
Both of these issues can be raised as separate tickets, however - I'm happy to unblock by accepting this now.  Great work!

jamesrwelch:
Other issues created as #607 and #608

PR #602, opened 2022-07-08
gh-590 Change architecture of bulk editor

### LINKED ISSUE #590: Create profile abstraction layer for bulk editor
# Overview

To support the bulk editor epic task #578, there needs to be some architectural changes to the bulk editor as it stands.

Currently all bulk editor UI components directly use the Mauro API (`mdm-resources`) to get and modify profiles. However, by introducing the concept of a "default" profile (and equivalent profile provider service which does not actually exist in the Mauro backend), this will introduce a lot of condition code in the form of "if default profile ... else ..."

The first step is to refactor the bulk editor and introduce an abstraction layer that sits between the UI components and the Mauro API:

Bulk edit UI ideas-Code architecture-2 drawio

All interactions should go through the abstraction layer. In the future task #592, this will then be extended to adjust the abstraction layer to handle both local and remote profile operations.

# Goals

1. All UI components must request profiles via the new abstraction layer and _not_ directly reference the Mauro API
2. Support all existing operations the bulk editor uses profiles for.
3. Make it easier for future work on the bulk editor to make internal changes by removing the hard dependencies.

# Functionality

The abstraction layer must be able to support the following profile operations:

1. List all available profiles for a domain type
2. Get all profile objects required - `getMany`
3. Validate a collection of profile objects - `validateMany`
4. Update a collection of profile objects - `saveMany`

pjmonks:
* Create new service layer between components and Mauro REST API

Also added general fixes to bulk editor

* Remove mix and match of Angular forms and reactive forms binding
* Fix template error when parent data model not loaded yet
* Fetch all data elements, not just first 20

Resolves #590

### DIFF
--- a/src/app/bulk-edit/bulk-edit-select/bulk-edit-select.component.ts
+++ b/src/app/bulk-edit/bulk-edit-select/bulk-edit-select.component.ts
@@ -16,21 +16,34 @@ limitations under the License.
 
 SPDX-License-Identifier: Apache-2.0
 */
-import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
+import {
+  Component,
+  EventEmitter,
+  Input,
+  OnDestroy,
+  OnInit,
+  Output
+} from '@angular/core';
 import { FormControl, FormGroup, Validators } from '@angular/forms';
-import { CatalogueItem, DataElement, DataElementIndexResponse, ProfileSummary, ProfileSummaryIndexResponse } from '@maurodatamapper/mdm-resources';
+import {
+  CatalogueItem,
+  DataElement,
+  DataElementIndexResponse,
+  ProfileSummary
+} from '@maurodatamapper/mdm-resources';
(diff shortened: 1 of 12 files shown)


src/app/bulk-edit/bulk-edit-select/bulk-edit-select.component.ts
```diff
@@ -42,6 +55,8 @@ export class BulkEditSelectComponent implements OnInit {
   availableProfiles: ProfileSummary[];
   setupForm: FormGroup;
 
+  private unsubscribe$ = new Subject();
```
OButlerOcc, line 58: can you explain this to me if you have time, its seems useful and I think i get whats its doing but I would like to know how its working.

OButlerOcc (approved):
Lots of formatting changes assume this was old file.
Looks fine as I can see. New services are good.
If you have time, would like an explanation on how you've used unsubscribe$, seems useful.

PR #603, opened 2022-07-12
Fixed tests and context bug

OButlerOcc:
Changes to match the expected context parameters
Updated tests to include new params


pjmonks (approved):
This looks better, thanks ✔️

Issue #592, opened 2022-07-06
Integrate the default profile provider service into the bulk editor


pjmonks:
**Note:** this depends on:

* #590
* #591 

# Overview

The bulk editor will be updated to fetch profiles from two sources:

1. The default profile provider service - this is designed to return only locally defined `Profile` objects in the UI that "pretend" they come from Mauro
2. The Mauro profile provider service - which uses the Mauro APIs to manage profiles (as expected)

By combining the two together, this means the default `Profile` can be applied to a bulk editor tab just like any other profile.

Issue #576, opened 2022-06-30
Next iteration of spreadsheet-style bulk editor


jamesrwelch:
The first iteration of the bulk editor tool provides a grid-based view to edit the profile fields of selected Data Elements from a Data Model.  This Epic defines some tasks for the next iteration of this functionality - improving its usability and making it more applicable across the rest of Mauro.  There's one ticket for each task, but these may need to be broken down into sub-epics if necessary

# Tasks

- [x] #577
- [x] #578
- [x] #579
- [x] #580
- [x] #581
- [x] #582
- [x] #583
- [x] #584
- [x] #585

# Related

- MauroDataMapper/docs#63

pjmonks:
All tasks have been completed in this epic now.

Issue #69, opened 2021-04-29
Unable to edit elements or data in reference data types


gammonpeter:
**Describe the bug**
Unable edit elements or data in reference data types

pjmonks:
Reference data types and elements have their UIs updated to support add/edit/delete.

Reference data values UI has started to get updated, however it was discovered that the backend does not yet support all the necessary endpoints for Reference Data Values. Submitted this ticket to YouTrack: https://metadatacatalogue.myjetbrains.com/youtrack/issue/MC-9497.

This issue is now on hold until the backend has been updated.

pjmonks:
I recently restarted work on this after seeing that `mdm-core` now should have the endpoints required to correctly manage all objects related to Reference Data Models - elements, types, data etc.

However, I have put it on hold again because I think the user design/experience should be determined first. For example, my original intent was to add new elements/types/data inline in the table views of each tab, but think perhaps that it would be better to use a wizard style interface, like those used for the Data Model/Class/Element/Type domains. This needs to be discussed though to agree on the correct approach.

Also, there are also YouTrack tickets also related to this area, and having the same issue listed in two different issue trackers is going to cause confusion. See https://metadatacatalogue.myjetbrains.com/youtrack/issue/MC-9486.

When this is revisited:

1. Decide which take priority - the YouTrack ticket or this GitHub issue.
2. Plan out the UX for managing/editing Reference Data Models.

pjmonks:
@jamesrwelch I remember working on this originally last year but pausing it for the reasons I've listed above. I feel that Reference Data Models should probably be setup as a new set of issues to properly plan/design their usage in the UI. Then this particular issue would be closed as it would be obsolete.

Would you agree?

jamesrwelch:
@pjmonks Yes, completely agree - good plan 👍

PR #604, opened 2022-07-15
gh-951 Create DefaultProfileProviderService

### LINKED ISSUE #591: Implement a default profile provider interception service
# Overview

Each model catalogue item contains some hard properties that are always included as metadata for the model, such as `description`, `author`, `organisation`, `aliases` etc. After that, any additional metadata can be added to the item, and some of this metadata can be grouped into "profiles" based on relevance.

It is now useful to treat the standard catalogue item properties as its own profile, the "default" profile. Mauro does not have a profile provider to group together fields like `description`, so an interception service will be created in the UI to manage this instead.

# Default profile

The default profile will actually be a manually created `Profile` object that is returned and managed, mapped to a catalogue item. It will contain one section, which will then contain the standard catalogue item fields - `description` for example. Each profile field will be fully defined in code, so a `description` will have a:

1. Display name
2. Metadata property name
3. Data type
4. Validity settings
5. etc

# Getting a default profile

Normally you request a `Profile` object in conjunction with a catalogue item e.g. a Data Model. This provider service will handle that by:

1. Manually defining the `Profile` object
2. Adding necessary profile fields - note that some domain types have some alterations to their standard fields, e.g. terms.
3. Mapping the catalogue item field values to the `currentValue` properties of each profile field
4. Then return the `Profile` object.

# Validation

In the case of a default `Profile`, no fields should be mandatory, so no additional validation error messages need to be applied to the `Profile` - simply return the object provided.

# Updating the default profile

When it is time to save changes made to the default `Profile`, this will not use the Mauro profile resources, but instead:

1. Map the `currentValue` properties of each profile field back to the catalogue item in question
2. Perform a direct `POST` request on that catalogue item endpoint to save changes

# Singular vs Many

The default profile provider service needs to support both management of single catalogue items and handling multiple catalogue items as once. For multiple catalogue items, the same principles of mapping object to/from each other still apply.

# Observables

Although some of these operations do not require asynchronous interactions (e.g. via a network request), they should still return `Observables`. This can then match a common interface with the original Mauro REST API.

pjmonks:
Interception service responsible for mapping catalogue item hardwired properties to a `Profile` object and back. This then allows the profile to be applied to other UI controls that require only a profile.

* Map original Mauro profile operations so that a in-memory profile object can be managed
* Create additional helper services to get/set any kind of Mauro item using only minimal identifier information. This circumvents the usual Mauro APIs which require you to know the domain you're working in upfront
* Full test suite provided

Resolves #591 

**Note:** the `DefaultProfileProviderService` isn't injected anywhere at the moment. Once approved, this is planned to be integrated into the bulk editor in issue #592

### DIFF
--- a/src/app/mauro/mauro-item-provider.service.spec.ts
+++ b/src/app/mauro/mauro-item-provider.service.spec.ts
@@ -1,0 +1,493 @@
+/*
+Copyright 2020-2022 University of Oxford
+and Health and Social Care Information Centre, also known as NHS Digital
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+import { CatalogueItemDomainType, Uuid } from '@maurodatamapper/mdm-resources';
+import { MdmResourcesService } from '@mdm/modules/resources';
+import { setupTestModuleForService } from '@mdm/testing/testing.helpers';
+import { cold } from 'jest-marbles';
+import { Observable } from 'rxjs';
+import { MauroItemProviderService } from './mauro-item-provider.service';
+import {
+  MauroIdentifier,
+  MauroItem,
+  MauroItemResponse
+} from './mauro-item.types';
+
+describe('MauroItemProviderService', () => {
+  let service: MauroItemProviderService;
+
+  const resourcesStub = {
+    dataModel: {
+      get: jest.fn() as jest.MockedFunction<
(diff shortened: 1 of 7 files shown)


pjmonks:
@aaronforshaw Not yet, I probably should have mentioned that in my original PR description. There are a couple of reasons why I wanted to hold off on those:

1. #576 says that the bulk editor should be improved to include all domain types for Data Models, Terminologies and Code Sets, but Reference Data Models are not included in that yet. As this change is driven by that epic, I wanted to only focus on the essentials first.
2. Issue #69 also discusses updates to the Reference Data Model UI for viewing/editing those domain types, so it kind of feels that Reference Data Models isn't fully supported in `mdm-ui` fully yet anyway.

Reference Data Models should be included in the future, but that can be a separate issue.

aaronforshaw:
OK 👍

aaronforshaw (approved):
Looks good. From memory I think mdm-ui and mdm-explorer have data structures similar to the new MauroIdentifier (although probably they are Data Model specific). Maybe at some point these could be consolidated.

pjmonks:
@aaronforshaw You're right, in fact there are a lot of components and structures that can be shared between UI projects. I had looked in the past at Angular component libraries to try and create shared reusable components, but hit some roadblocks. It is something that should be reviewed again in the future.

Thanks for the approval 👍

PR #606, opened 2022-07-15
Feature/mc 9809

jamesrwelch:
Simplified editor for creating new data types whilst creating a new data element

### DIFF
--- a/src/app/utility/new-data-type-inline/new-data-type-inline.component.html
+++ b/src/app/utility/new-data-type-inline/new-data-type-inline.component.html
@@ -27,43 +27,18 @@ SPDX-License-Identifier: Apache-2.0
 					<mdm-model-path [path]="[parentDataModel]"></mdm-model-path>
 				</td>
 			</tr>
-			<tr>
-				<td class="detailsRowHeader">Label *</td>
-				<td>
-					<mat-form-field appearance="outline" class="full-width">
-						<input matInput name="label" [(ngModel)]="model.label" required>
-					</mat-form-field>
-				</td>
-
-			</tr>
-			<tr>
-				<td class="detailsRowHeader">Description</td>
-				<td>
-          <mdm-content-editor
-            [(content)]="model.description"
-            [inEditMode]="true">
-          </mdm-content-editor>
-				</td>
-			</tr>
-			<tr *ngIf="showClassification">
-				<td class="detailsRowHeader">Classifications</td>
-				<td>
-					<mdm-element-classifications [readOnly]="false" name="classifiers" (classificationsChanged)="model.classifiers = $event">
-					</mdm-element-classifications>
-				</td>
-			</tr>
-			<tr class="type" style="border-top: 0px;">
-				<td class="detailsRowHeader">Type *</td>
-				<td>
-					<div class="mdm--form-input">
-						<mat-form-field appearance="outline" class="full-width">
-							<mat-select name="dataModelType"  [(ngModel)]="model.domainType" required (ngModelChange)="onTypeSelect()">
-                        <mat-option [value]="dt.id" *ngFor="let dt of allDataTypes" [hidden]="dt.displayLabel === 'ModelDataType'">{{dt.displayLabel}}</mat-option>
-							</mat-select>
-						</mat-form-field>
-					</div>
-				</td>
-			</tr>
+      <tr class="type" style="border-top: 0px;">
+        <td class="detailsRowHeader">Type *</td>
+        <td>
+          <div class="mdm--form-input">
+            <mat-form-field appearance="outline" class="full-width">
+              <mat-select name="dataModelType"  [(ngModel)]="model.domainType" required (ngModelChange)="onTypeSelect()">
+                <mat-option [value]="dt.id" *ngFor="let dt of allDataTypes" [hidden]="dt.displayLabel === 'ModelDataType'">{{dt.displayLabel}}</mat-option>
+              </mat-select>
+            </mat-form-field>
+          </div>
+        </td>
+      </tr>
 			<tr class="type" *ngIf="model.domainType != 'PrimitiveType'">
 				<td class="detailsRowHeader">
 					<span *ngIf="model.domainType === 'EnumerationType'">Enumeration (key,value) *</span>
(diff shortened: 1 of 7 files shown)


jamesrwelch:
As discussed, the new data type inline form needs checking to ensure it hasn't broken any of the other places where it appears.

src/app/utility/new-data-type-inline/new-data-type-inline.component.ts
```diff
@@ -70,6 +72,7 @@ export class NewDataTypeInlineComponent implements OnInit, AfterViewInit {
 
   ngOnInit() {
     this.validate();
+    console.log(this.parentDataModel);
```
pjmonks, line 75: Once you're happy with how it works and tested, this `console.log()` line can be removed.

src/app/utility/new-data-type-inline/new-data-type-inline.component.ts
```diff
@@ -188,4 +191,8 @@ export class NewDataTypeInlineComponent implements OnInit, AfterViewInit {
     this.validate();
     this.sendValidationStatus();
   };
+
+  toggleAdvanced(): void  {
+    this.advanced = !this.advanced;
```
pjmonks, line 196: Could you update `onTypeSelect()` to set `advanced = false` please? I've identified an error where the `advanced` state is not reset correctly, as follows:

1. Change type from `Primitive` to `Reference`
2. Toggle the Advanced Options to see further fields
3. Now change type back to `Primitive`
4. And change back again to `Reference`. Note that the "Advanced options" toggle is in the off state, but the advanced options are visible. Now trying to toggle has the opposite effect.

You'll also need to bind the `advanced` field to the `mdm-slide-toggle` component to keep track of initial state correctly.

src/app/utility/new-data-type-inline/new-data-type-inline.component.html
```diff
@@ -160,7 +135,37 @@
...
+      <tr *ngIf="showLabelAndDescription || model.domainType === 'PrimitiveType' || model.domainType === 'EnumerationType' || this.advanced">
+        <td class="detailsRowHeader">Data Type Label *</td>
+        <td>
+          <mat-form-field appearance="outline" class="full-width">
+            <input matInput name="label" [(ngModel)]="model.label" required>
+          </mat-form-field>
+        </td>
+      </tr>
+      <tr *ngIf="showLabelAndDescription || model.domainType === 'PrimitiveType' || model.domainType === 'EnumerationType' || this.advanced">
+        <td class="detailsRowHeader">Description</td>
```
pjmonks, line 152: Why is the "Description" field considered an advanced option for Reference, Terminology, Code Set and Reference Data Model types but not others? Maybe there is a reason, but from me just looking at the form it confuses me.

src/app/utility/new-data-type-inline/new-data-type-inline.component.ts
```diff
@@ -70,6 +72,7 @@ export class NewDataTypeInlineComponent implements OnInit, AfterViewInit {
 
   ngOnInit() {
```
pjmonks, line 73: I've identified a potential memory leak. This component does not have an `ngOnDestroy` function to clean up the subscription created in `ngAfterViewInit`:

```ts
this.formDataTypeChangesSubscription = this.myFormNewDataType.form.valueChanges.subscribe(
      x => {
        this.validate(x);
      }
    );
```

If this is not unsubscribed then the subscription is kept in memory. This should be cleaned up by:

1. Implementing the `OnDestroy` Angular interface on the component
2. Either:
    1. Calling `this.formDataTypeChangesSubscription.unubscribe()` in `ngOnDestroy`, or
    2. Use the declarative version of unsubscribe: https://www.digitalocean.com/community/tutorials/angular-takeuntil-rxjs-unsubscribe

pjmonks:
There are 3 areas where the inline Data Type creator component is used:

1. Data Element wizard (step 2)
2. Data Type wizard (step 2)
3. Default profile editor - i.e. choose to edit and existing Data Element or Data Type

I've tested on all 3 and the form layout seems fine, though there is a lot of fields jumping about depending on a few factors e.g. which data type, is advanced options picked etc. It just looks like quite a complex set of fields to define depending on what you want to do.

One issue I found in all 3 though is lack of overall validation. In the wizards I can try to create a new Data Element without filling in the new data type details, but see no validation messages. When editing an existing Data Element, I can try and setup part of a new data type but miss required fields, then click Save and watch it attempt to submit but then return a vague Mauro endpoint error message back.

pjmonks (changes requested):
I like the new table layout for the wizard and inline control forms you're going for. I have spotted some issues though, both in the code and functionality, which I've highlighted in my review if could look at them please.

src/app/utility/new-data-type-inline/new-data-type-inline.component.ts
jamesrwelch: Yep, good spot

src/app/utility/new-data-type-inline/new-data-type-inline.component.ts
jamesrwelch: Oh - yes, brilliant, thanks.  Much neater.

src/app/utility/new-data-type-inline/new-data-type-inline.component.html
jamesrwelch: For the data types where we reference something else, the back-end can now supply a suitable data type name - e.g. "Reference to X".  That's what we tend to enter manually anyway.  And the description we'd leave blank - it's just a reference to something.  For primitives and enumerations, you're creating something new, and so you need to choose a name, at least, and ideally a description.

src/app/utility/new-data-type-inline/new-data-type-inline.component.ts
jamesrwelch: Yes, again, good spot.  I've gone with the first option - the declarative version looks cleaner, but doesn't seem to save much code, and I couldn't see it used elsewhere, so keeping things consistent.

jamesrwelch:
@pjmonks Thanks for the detailed review comments - I've addressed the issues and also fixed an issue with the validation logic I noticed when re-testing.

pjmonks (approved):
Thanks for addressing the feedback @jamesrwelch , it's looking a lot better. My previous comments on the overall UX of the component can become new issues if necessary, best to let the users try it out first and see what they think.

Nice one 👍

PR #609, opened 2022-07-18
Added hover description tooltips to col names

OButlerOcc:


pjmonks (approved):
Great job. I made a minor change to not include a default tooltip if there was no description in a profile field, but the rest is good. ✔️

Issue #71, opened 2021-04-29
Suggested Links window fails when finding links


gammonpeter:
**Describe the bug**
On a Data Model go to the links view and click on suggested links, select and target and click the suggest links button a 404 is returned

gammonpeter:
Fixed with BE update

PR #612, opened 2022-07-20
MC-9891 Update subscribed catalogue UI due to breaking core changes

pjmonks:
**Note:** this depends on MauroDataMapper/mdm-resources#71

* Update admin setup forms to use Angular reactive forms
* Add new connection type field to add/edit subscribed catalogue
* Display connection type of each subscribed catalogue in the list view
* Handle case where published model type is not known
* Display published data model dates
* Update Subscribe Data Model dialog to use Angular reactive forms
* Send additional properties during subscribed model creation - content link and/or importer

# Screenshots

Admin screen of subscribed catalogues with subscribed catalogue types:

image

Inclusion of dates from published data models:

image

Updates to subscribe dialog:

image

### DIFF
--- a/src/app/admin/subscribed-catalogue/subscribed-catalogue.component.html
+++ b/src/app/admin/subscribed-catalogue/subscribed-catalogue.component.html
@@ -17,49 +17,84 @@ limitations under the License.
 SPDX-License-Identifier: Apache-2.0
 -->
 <div class="heading-container mb-2 forms-heading-container">
-    <h4>
-        <span *ngIf="catalogue.id">Edit Subscribed Catalogue</span>
-        <span *ngIf="!catalogue.id">Add Subscribed Catalogue</span>
-    </h4>
+  <h4>
+    <span *ngIf="catalogueId">Edit Subscribed Catalogue</span>
+    <span *ngIf="!catalogueId">Add Subscribed Catalogue</span>
+  </h4>
 </div>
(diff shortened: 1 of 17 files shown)


joe-crawford (changes requested):
Thanks for the PR @pjmonks.

I'm testing this by using `develop` of mdm-core, `feature/mc-9681` of mdm-ui and mdm-resources.

I have been able to use the UI to add a Mauro JSON subscription to https://modelcatalogue.cs.ox.ac.uk/continuous-deployment and use the new modal to subscribe to a model.

There's an issue when viewing published models when connecting to an Atom catalogue however, I'm testing using an Atom subscription to https://modelcatalogue.cs.ox.ac.uk/continuous-deployment/api/feeds/all (with an API key).

I can see a list of published models in the tree on the left hand side of the UI, however when selecting these models a  Not Found error appears in the main panel. Could you have a look at this?

N.B. When subscribing to an Atom catalogue, the `modelId`s returned in the `/publishedModels` and other endpoints will be unique but are not of any particular format as they come from the Atom source. In the Mauro Atom subscribed catalogue they are of the form `urn:uuid:<UUID>` but in general they could be any string. I'm wondering if this may be causing the issue.

joe-crawford:
I've noticed the backend error text isn't very user friendly when selecting an incompatible model content type and importer to subscribe to, I'll try to update this.

joe-crawford:
Also looking at the published models ordering on the backend as that seems out of order.

PR #616, opened 2022-07-22
gh-592 Integrate default profile provider into bulk editor

### LINKED ISSUE #592: Integrate the default profile provider service into the bulk editor

pjmonks:
Allows the default, standard catalogue item properties (like "description") to be mass edited alongside other profiles.

Resolves #592 

* Make default profile an option in bulk editor
* Define common interface between profile provider services
* Bulk editor supports loading default profile objects
* Update bulk edit profile validation support
* Catch Mauro validation errors from HTTP 422 response and pass down the observable pipeline as a regular response object. This then consolidates the error counting logic downstream in the component
* Update bulk edit support for saving
* Adjust UI so that validate/save controls are within each tab. Every profile tab is now in control of those operations
* Fix validation of multiplicity fields in default profile
* Profile fields of model type are readonly and display label now
* Updated test suite

# Screenshots

"Default profile" is now an option to select

image

Displaying the "default" profile of a list of Data Elements. Notes that the save, validation and validity components are now in a toolbar above the spreadsheet.

image

Example of invalid fields in new UI layout

image


OButlerOcc (approved):
so if I went through this correctly.
There's a service that provides a default profile. 
This profile is included in the list of available profiles for a mauro item (is a mauroitem a catch all for any catalogue item but extends the definition to include a few extra bits of data?) Seems so looking at it.
You can select the default profile from the list when selecting profiles for items and can edit and save it.

It looks sensible to me, in the context of what a default profile is. Its a little weird editing and saving a default to "undefault?" it rather than just persisting as a default profile but I think thats in line for scope for default profiles.

Issue #618, opened 2022-07-25
Mark as Deleted Data Models are not visible to Administrators


GBishop-PHSA:
# Description
For Data Models that are created in sub-folders, when the Show Deleted Models filter is checked, Mark as Deleted data models do not appear in the Tree.

# Steps to reproduce
1. Login as administrator
2. Find a data model that is in a sub-folder in the browse tree
3. Mark the data model as deleted
4. Use the filter in the Browse Tree to Include Deleted models or set user preferences to see Deleted data models (User>Preferences>Show Deleted Models)
5. Refresh page the Browse Tree

# Expected behavior
The data model appears in the sub folder with Strikethrough text

# Actual behaviour
The deleted data model does not appear in the sub folder

# Environment

## Desktop 

- OS: Windows 10
- Browser: Chrome
- Version: 103.0.5060.114 

# Additional context
The expected behaviour may have changed due to this update to mdm-core: https://github.com/MauroDataMapper/mdm-core/commit/21ec326adc1f8a720542bde452312b70cb08d0c1#diff-56322d760db078f3c89b053994f54fdff24ee21cf2a735596f92f1a0c74c50cf

pjmonks:
This appears to be fixed in the 7.3.0 release.

With option to see deleted models (as an administrator):

![image](https://user-images.githubusercontent.com/3219480/221885264-114d209c-5e53-472e-91f9-5258e7722f84.png)

With option turned off:

![image](https://user-images.githubusercontent.com/3219480/221885459-4d4045ed-7423-41d0-ae23-913cf2defca4.png)

Closing this issue as a result.

PR #619, opened 2022-07-25
gh-618 Show mark as deleted data models to admins

GBishop-PHSA:
When an admin user selects the Show Deleted Models filter option, the data models that are marked as deleted should show in the browse tree with a strikethrough font.


GBishop-PHSA:
@joe-crawford  This is ready for review

joe-crawford:
Thanks @GBishop-PHSA, works well locally.

joe-crawford (approved):

Issue #595, opened 2022-07-06
Update bulk edit tabbed UI to support adding/removing profiles


pjmonks:
# Overview

The bulk editor currently makes the user select a list of profiles in one screen before displaying a tabbed interface on the next. This enhancement suggest moving the profile selection to the same tab interface.

# Design

![Bulk edit UI ideas-Profile selection drawio](https://user-images.githubusercontent.com/3219480/177528173-1a1e1ef6-966b-4b84-abbd-575214406c73.png)

# Default profile tab

The default profile tab will always be present and cannot be removed, allowing the user to bulk edit the standard item properties.

**Note:** this will require #578 to be completed first.

# Adding tabs

A new "+" button will be added to the tab interface to add more profiles. Clicking the "+" button will display a popup menu of available profiles to use (not including the ones already in tabs).

Selecting a profile will render a new tab tied to that profile, listing the same child items selected in the previous page but mapped to this profile.

# Removing tabs

Apart from the default profile tab, all other tabs should be able to be removed. If there are any unsaved edits in the tab though,  confirm with the user first whether to close the tab and lose information.

pjmonks:
Whilst working on #592 I've redesigned the spreadsheet tabs a bit already:

![image](https://user-images.githubusercontent.com/3219480/180204259-224e1de1-9200-4d6f-babe-c491dfc534ca.png)

Looking at this design and comparing with the feature set above, these might also be nice things to have (but optional depending on time):

1. Show if cells are "dirty" changed
2. Disable validate/save buttons if no changes have been made yet
3. Count how many cells were changed so far (display next to validation message)

pjmonks:
After thinking about this and the changes made in #594 , perhaps the UI is in a stable enough state now. Selecting the profiles to use does not seem onerous, and any further additions may end up cluttering the UI more. The additional bonus suggestions also do not seem as worthwhile as I thought.

I think this issue should be closed and we wait for real-world user feedback to drive further requirements forward.

PR #621, opened 2022-07-27
gh-594 Update bulk editor item selection

### LINKED ISSUE #582: Make the bulk editor work at the Data Class level
At the moment, you can only bulk-edit from the Data Model Level.  It would be nice if you could do the same thing at the DC level - to select from the DEs contained within that DC (and its contained classes).
This may reduce the priority of #577 .

### LINKED ISSUE #583: Make the bulk-editor tool work for terms in a terminology or a codeset
At the moment, the bulk editor only works for DEs within a DM.  It would be nice if we can get it working for Terms within a Terminology, and within a CodeSet

### LINKED ISSUE #584: Make the bulk-editor work on data classes, data types.
At the moment the bulk editor only works for DEs within a DC.  It would be nice if we could also edit DCs or DTs within a DM (perhaps in separate screens, rather than all together, if that's an easier starting point).

### LINKED ISSUE #594: Update child item selection for bulk editor

pjmonks:
Resolves #594
Resolves #582 
Resolves #583 
Resolves #584 

* Add child domain type selection to pick more than Data Elements
* Change child item selection UI to use detailed selection list
* Move breadcrumb component out of catalogue search and into shared module
* Restyle the select form
* Change buttons in spreadsheet view to make more sense
* Clear context when going back to item selection
* Support correct root item when bulk editing Data Classes

Also resolves gh-582, gh-583 and gh-584 to allow bulk editor to work with more domain types:

* Terminologies
* Code Sets
* Data Classes

**Note:** the selection of profiles is still inside the "selection list" area of the bulk editor. Issue #595 will move that to the tabbed spreadsheet view.

# Screenshots

Select different child domain types:

image

Updated item selection list UI:

image

Example of Data Classes being bulk edited:

image


aaronforshaw:
This looks good and I found the bulk editing workflow much faster to use 👍 

A few usability comments:
- When bulk editing terms of a terminology (e.g. using the bootstrapped Complex Test Terminology) only the first 10 terms can be selected. I think this is because the endpoint `http://localhost:8080/api/terminologies/{terminologyId}/terms?all=true&sort=label` intentionally does not actually return all terms. So some pagination might be required so that a user can select more than just the first 10 terms.
- The left column of the bulk editing table is headed 'Elements', which makes me think of data elements. Possibly this could be headed 'Items' instead, given that the column can be for data elements, classes, terms etc....
- After successfully saving some edits, when I clicked on the 'Close' button there is a confirmation saying that any unsaved changes will be lost. In future this could be enhanced to check for any unsaved changes; if there are none then close without the confirmation modal.
- It took me ages to find the save button. Elsewhere save is done using a button labelled 'Save' or by pressing a green checkmark. It might be good to use one of these rather than using the floppy disk icon.
- The 'Domain type' select could display the domain types capitalised e.g. 'Data Type' rather than 'Data type' for consistency with other parts of the UI
- In the default profile I haven't been able to save classifiers or aliases. I don't know whether this is due to the bulk editing change, or the default profile change.

src/app/bulk-edit/bulk-edit-select/bulk-edit-select.component.ts
```diff
@@ -131,18 +188,111 @@ export class BulkEditSelectComponent implements OnInit, OnDestroy {
...
+  }
+
+  private getSuitableChildDomainTypes(): CatalogueItemDomainTypeOption[] {
+    if (
+      this.context.rootItem.domainType === CatalogueItemDomainType.DataModel
+    ) {
+      return [
+        {
+          domainType: CatalogueItemDomainType.ModelDataType,
+          displayName: 'Data types'
```
aaronforshaw, line 205: Capitalised as 'Data Types', 'Data Classes' and 'Data Elements' perhaps?

src/app/bulk-edit/bulk-edit-select/bulk-edit-select.component.ts
```diff
@@ -131,18 +188,111 @@ export class BulkEditSelectComponent implements OnInit, OnDestroy {
...
+    ) {
+      request$ = this.resources.dataType.list(
+        this.context.rootItem.id,
+        filters
       );
+    } else if (this.context.childDomainType === CatalogueItemDomainType.Term) {
+      request$ =
+        this.context.rootItem.domainType === CatalogueItemDomainType.CodeSet
+          ? this.resources.codeSet.terms(this.context.rootItem.id, filters)
+          : this.resources.term.list(this.context.rootItem.id, filters);
```
aaronforshaw, line 279: This might need paginating

src/app/bulk-edit/bulk-edit-select/bulk-edit-select.component.html
```diff
@@ -16,53 +16,120 @@
...
+      <p>
+        Select a domain type that contains child catalogue items to continue.
+      </p>
+    </div>
+    <div
+      *ngIf="availableChildItems.length > 0"
+      class="mdm-bulk-edit__selection"
+    >
+      <p>
+        Please select the elements from this model and profiles you wish to
```
aaronforshaw, line 57: 'items' rather than 'elements'

aaronforshaw (changes requested):

src/app/bulk-edit/bulk-edit-select/bulk-edit-select.component.ts
pjmonks: No problem, fixed

src/app/bulk-edit/bulk-edit-select/bulk-edit-select.component.ts
pjmonks: Need to look into this one, did not know that the `all` param doesn't work for term lists.

src/app/bulk-edit/bulk-edit-select/bulk-edit-select.component.html
pjmonks: Fixed this now

pjmonks:
Thanks for the feedback @aaronforshaw . Some of this was simple to change, very minor. The terms list I need to look into, I might be able to implement virtual scrolling in the selection list to act as pagination. I wasn't aware of that limitation for term lists.

Your other points I will investigate but may push out into separate issues:

1. Aliases/classifiers might need some custom cell editors. Classifiers at least I would imagine needs to update with IDs of classifiers rather than a string
2. Only seeing the "do you want to leave" confirmation would definitely make sense only if you haven't saved, it would just need t be tracked correctly. Each tab now saves it's own profile now.

src/app/bulk-edit/bulk-edit-select/bulk-edit-select.component.ts
pjmonks: I've temporarily worked around this by doing an initial list request to get the total count, then doing a second list request with `max: count`. It could be improved at a later date.

pjmonks:
@aaronforshaw I've completed all the feedback I can for now. I've created these separate issues to work on later:

* #623
* #624 
* #625

aaronforshaw (changes requested):
Thanks for addressing all the comments, raising new tickets where necessary.

I just noticed one other issue, which I didn't spot earlier: I think the validate button and indicator should be hidden for the default profile.

pjmonks:
@aaronforshaw There is actually one case when we validate the default profile, it checks if the multiplicity value you entered is in the correct format e.g. "0..1" and not some random numbers/characters. To adjust that properly that would probably also require a custom cell editor, which I think would be another issue to raise.

aaronforshaw (approved):

PR #626, opened 2022-07-28
Feature/gh 601 - hide not accepted items in search modal tree

### LINKED ISSUE #601: Model selector tree in catalogue search displays items that are not acceptable
# Description

The model tree selector modal in the catalogue search is configured to use only accept these domain types on selection:

```html
<mdm-model-selector-tree
      [ngModel]="context"
      [onSelect]="onContextSelected"
      ngDefaultControl
      [placeholder]="'Choose a Folder, Data Model or Data Class'"
      [accepts]="['Folder', 'DataModel', 'DataClass', 'ReferenceDataModel']"
      [multiple]="false"
      [alwaysShowTree]="true"
      [usedInModalDialogue]="true"
      [showInputGroup]="false"
      [doNotShowChildDataClasses]="true"
      >
    </mdm-model-selector-tree>
```

However, when you see the tree view, other domain types are visible, which appears confusing if they are not able to be selected:

image

Also, you are able to select these non-acceptable domain types in this way:

1. Try to select a Terminology - the "OK" button remains disabled
2. Now click a Data Model - the "OK" button is enabled, as expected
3. Click the Terminology again - the "OK" button is still enabled (not expected)
4. Click "OK". The model returned is the Data Model from step 2, not the Terminology

This sequence of steps appears confusing to the user - which was allowed to be selected? And if something else was selected, why was it not used?

# Expected behavior

Ideally the model tree should only be displaying acceptable selections to avoid confusion.

aaronforshaw:
Closes #601 

The `mdm-folders-tree` component already had an input property `filterByDomainType`, which, if present, removes items with a non matching domain type from the tree.

This change adds the equivalent input property to the `mdm-model-selector-tree` component, which is then passed into `mdm-folders-tree`. 

Then, in the `mdm-catalogue-item-select-modal` component, this property is set to only allow Data Model, Data Class, Reference Data Model and Terminology to appear in the tree.


pjmonks (approved):
Great! This looks more like what I was expecting ✔️

PR #629, opened 2022-07-29
gh-585 - Edit Markdown and HTML text in the bulk editor

### LINKED ISSUE #585: Enable the HTML / MD editor on text fields within the profile editor
We made a simplification in the first iteration of the bulk editor, to not try any fancy editing inside the boxes - in particular not using the Jodit / MD editor within the grid.  For space / rendering issues, I think having an HTML editor on each text box might be un-workable, but maybe there's a way to pop-up an HTML / MD editor on request?

pjmonks:
Resolves #585 

Updates the bulk editor to support full Markdown and HTML editing on any profile field that has the type `text` - the `description` field in the "Default profile" is a good example.

* Uses a basic popup cell editor for basic text edits within the cell, as expected for agGrid.
* There is then a "pop out" button to expand the cell value into a separate dialog. This gives full editor control over the Markdown/HTML text.

# Screenshots

image

image


src/app/bulk-edit/bulk-edit-editor/bulk-edit-editor.component.ts
```diff
@@ -208,10 +226,28 @@ export class BulkEditEditorComponent implements OnInit {
...
+    const nonStandardFields = this.nonStandardColWidths.map((c) => c.field);
+
     const columnIds = this.gridColumnApi
       .getAllColumns()
+      .filter((col) => !nonStandardFields.includes(col.getColDef().field))
       .map((col) => col.getColId());
+
     this.gridColumnApi.autoSizeColumns(columnIds);
+
+    this.nonStandardColWidths
```
OButlerOcc, line 238: I've got a method for non standard col resizable, if we need more of these may be worth structuring as something we can dynamically adjust via properties or something?

OButlerOcc (dismissed):
Easy to follow, this one. Works as far as I've poked it locally. I've noticed we're having to non standardise a few column options now we're using custom components I wonder if there's a smarter way to do that.

OButlerOcc (approved):
don't see any changes from last time after rebase

PR #650, opened 2022-09-02
MC-9621 Allow UI to make updates to profiles editable after finalisation

pjmonks:
* Show edit profile button for profiles that are editable after finalisation
* Make individual profile fields readonly under conditions: if profile field is uneditable, or if item is finalised but profile field is editable after finalisation it overrides the readonly state
* Display info messages when item is finalised to make it clear to users
* Allow certain profiles to be added after finalisation. Only profiles with the editableAfterFinalisation flag can still be added

# Screenshots

Item is finalised but can still edit parts of this profile:

image

image

Add further profiles to a finalised item, but only suitable profiles:

image

### DIFF
--- a/src/app/modals/add-profile-modal/add-profile-modal.component.html
+++ b/src/app/modals/add-profile-modal/add-profile-modal.component.html
@@ -16,38 +16,45 @@ limitations under the License.
 
 SPDX-License-Identifier: Apache-2.0
 -->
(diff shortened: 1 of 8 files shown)


joe-crawford:
This works great and the messages look good too. Thanks @pjmonks!

joe-crawford (approved):

PR #651, opened 2022-09-02
MC-9719 Fix profile validation issues

pjmonks:
* Use correct fieldTotal property for counting validation errors
* Don't validate the empty profile immediately on open, only when clicking the validate button

### DIFF
--- a/src/app/modals/edit-profile-modal/edit-profile-modal.component.html
+++ b/src/app/modals/edit-profile-modal/edit-profile-modal.component.html
@@ -205,16 +205,16 @@ SPDX-License-Identifier: Apache-2.0
 </mat-dialog-content>
 <mat-dialog-actions align="end" class="pt-2 pb-2">
   <div>
-    <span *ngIf="validationErrors.total === 0" class="mr-2">
+    <span *ngIf="isValidated && validationErrors.total === 0" class="mr-2">
       <span class="fas fa-check-circle success"></span>
       Profile is valid
     </span>
-    <span *ngIf="validationErrors.total > 0" class="mr-2">
+    <span *ngIf="isValidated && validationErrors.fieldTotal > 0" class="mr-2">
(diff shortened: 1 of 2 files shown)


joe-crawford:
Thanks @pjmonks.

Looking at how this works, perhaps something more is needed. Not validating the (possibly new/empty) profile on opening is to avoid the user immediately seeing red 'field required' errors for the blank fields. But as validation is optional, now they might click Save and miss that a field is required if they don't spot the `*`.

Perhaps we could make an additional change, so the user is required to click Validate at least once before clicking Save. So Save is grey to start with, then is un-greyed on validation, even if it fails. 

@pjmonks, @jamesrwelch Do you think the above makes sense, or is that too many clicks?

jamesrwelch:
@pjmonks , @joe-crawford 

This is a difficult one.  We don't want to force people to validate the entry, but helping them create valid entries might be useful.  How about:
- We don't validate unless the validate button is clicked.
- The user can click 'save' at any point.
- When the user clicks save, we run a validation.  If the form entry is invalid, we ask pop up a box telling the user that the values are invalid - would they like to continue, or edit their response?  If they wish to save without validating, we continue; if they wish to continue editing we give them the hints as though they'd clicked the validate button.

joe-crawford:
Thanks @pjmonks and @jamesrwelch. I suppose the solution is either as James suggests, or perhaps I'm overcomplicating it in which case keeping the PR as it is might be best. Will leave to you to decide.

pjmonks:
@jamesrwelch , @joe-crawford I've updated the profile dialog as James has suggested:

image

If validation errors are found when saving, the user gets a chance to confirm whether to actually continue or correct the issues found.

joe-crawford (approved):
@pjmonks, sorry for the delay in getting to this. Looks good, thank you!

PR #652, opened 2022-09-02
gh-492 Display version and branch name of terminology when selecting for Code Sets

### LINKED ISSUE #492: List Version no. alongside terminology name when creating a new codeset
When you create a new codeset, you choose which terminology to choose terms from.  Where the terminology is versioned, or there are multiple versions in different versioned folders, it's hard to know which is which.  The list should show the version number to help ensure people pick the correct one.

pjmonks:
Resolves #492 

image

### DIFF
--- a/src/app/shared/element-label/element-label.component.html
+++ b/src/app/shared/element-label/element-label.component.html
@@ -1,0 +1,30 @@
+<!--
+Copyright 2020-2022 University of Oxford
+and Health and Social Care Information Centre, also known as NHS Digital
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


jamesrwelch:
This is brilliant - could we roll this out easily to other places too - for example when we create model data types?  Is there an easy way to find other similar model drop-downs?

pjmonks:
@jamesrwelch I've updated other areas which use a similar control:

image

I've updated the New Data Type wizard and the Element Selector (links for description fields).

Now that the label with branch/version is a shared component I've also looked at the tree views too. The new Favourites tree is updated but not the main model tree - too complex to fiddle with at the moment.

jamesrwelch (approved):
This works really nicely - thank you!

Issue #516, opened 2022-04-28
Terminology authoring doesn't allow adding a relationship between existing terms


oughnic:
# Description

When creating a terminology or editing a terminology in the Mauro Data mapper UI it is not possible to pick the target code for a relationship between two codes

# Steps to reproduce

1. Create a new terminology
2. Add a relationship type
3. Add a set of codes
4. Seek to add a relationship between two existing codes

# Screenshots

<img width="381" alt="Screenshot 2022-04-28 at 13 49 38" src="https://user-images.githubusercontent.com/27196895/165756106-62951ed5-41b1-49ab-8990-838e53df3065.png">

PR #658, opened 2022-09-08
gh-516 Solved duplicate Code check by showing an error

OButlerOcc:
Bug was caused to no check for duplicate codes in a term, which are forced to be unique. Added a check for this and display an error when a user attempts to create a new term with a duplicate code, so the term code can be ammended in window. Tested creating new and using existing terms for relationships. Auto Tests all pass.


OButlerOcc:
Superseded by https://github.com/MauroDataMapper/mdm-ui/pull/659

PR #659, opened 2022-09-08
gh-516 - Duplicate check added

OButlerOcc:
Bug was caused to no check for duplicate codes in a term, which are forced to be unique. Added a check for this and display an error when a user attempts to create a new term with a duplicate code, so the term code can be amended in window. Tested creating new and using existing terms for relationships. Auto Tests all pass.


pjmonks (approved):
I'm happy with this and find it works well. As I was looking at this and realising how the Angular `FormGroup` is used, it occurred to me that an alternative approach could have been to add a custom validator to the code form field to do the duplicate check when a value was entered - forcing the field to be invalid would not allow you to submit the form at all. What happens here though is perfectly fine.

One other minor comment is that I've noticed a couple of code comments describing the code, generally if the variables and functions are named well then the comments shouldn't really be needed, reading the code should be enough. Comments should only be used if the code is hard to follow and needs further explanation. Nothing to prevent this PR from being accepted though, feedback for the future.

Otherwise, this is approved ✔️

Issue #662, opened 2022-09-14
Allow user to view Drag and Drop sort order on screen entry of Data Standard elements


GBishop-PHSA:
# Description

When creating Data Standards, the order of the data elements in each class is important for data entry for profile properties. While users can drag and drop the ordering of data elements with a Data Standard, it's not clear to a user when entering the screen what the drag and drop order is. The UI should be updated to show the user the drag and drop order through a new column in the grid view.

# Additional context

The new column in the Element tab of the Data Class details screen will be called 'Display Order'
Users can click the column header to see what the display order is set to
The Display Order column is only visible for Classes that are a child of a Data Standard and are currently being edited

GBishop-PHSA:
Pull requests ready for review:
https://github.com/MauroDataMapper/mdm-ui/pull/666
https://github.com/MauroDataMapper/mdm-core/pull/365

PR #666, opened 2022-09-28
gh-662 New Display Order column for data element grid view

GBishop-PHSA:
gh-662 Show the disploy order column for data elements so the user can view how custom profile properties will appear on Data Model details screens.

### DIFF
--- a/src/app/shared/data-elements-list/data-elements-list.component.html
+++ b/src/app/shared/data-elements-list/data-elements-list.component.html
@@ -237,6 +237,14 @@ SPDX-License-Identifier: Apache-2.0
       </td>
     </ng-container>
     <tr mat-header-row *matHeaderRowDef="displayedColumns"></tr>
+    <ng-template [ngIf]="isOrderedDataSource">
+    <tr
+      mat-row
+      *matRowDef="let row; columns: displayedColumns"
+      [ngClass]="{ 'mdm--active-row ': row.checked }"
+    ></tr>
+    </ng-template>
+    <ng-template [ngIf]="!isOrderedDataSource">
     <tr
       mat-row
       *matRowDef="let row; columns: displayedColumns"
@@ -244,6 +252,7 @@ SPDX-License-Identifier: Apache-2.0
       [cdkDragData]="row"
       [ngClass]="{ 'mdm--active-row ': row.checked }"
     ></tr>
+    </ng-template>
   </table>
 </div>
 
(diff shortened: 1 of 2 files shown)


jamesrwelch:
@GBishop-PHSA I'm going to talk to @joe-crawford about this one, because he's also been thinking about the ordering of things (and made some changes to the back-end too).  

Are there circumstances where you'd want the elements to appear in anything other than index order?  My feeling is that we should have a flag on the class that says whether it should be ordered or not, and where it is, we only show elements in the precribed order, and where it isn't, we allow people to order by name, type, etc.  (For all Data Models, not just Data Standards.). That's a larger change though... and maybe this PR serves the purpose for the time-being.

GBishop-PHSA:
@jamesrwelch For our use case we would always want the elements to appear in index order. There just wasn't a way for users to confirm that the elements were sorted by index, without a column showing the numeric ordering.

joe-crawford:
Hi @GBishop-PHSA, thanks for this, although as James mentioned we'd previously thought of the index in the database as being a system property that isn't shown to the users. One issue is that in the case of imported data elements, the idx is actually the order in the original data class, because of that the backend doesn't actually use the index ordering in this case. Screenshot below:
Screenshot 2022-10-03 at 19 55 47

I do think there is a potential UI-only solution however. There are basically 3 states for each column, sort ascending, sort descending, don't sort, clicking on any column cycles through the 3 states. However it looks like the 3rd state doesn't work, as it still leaves the `sort=<column>` parameter in the request. If this was solved so that the UI did cycle through the 3 states properly, i.e. `sort=<column>&order=asc`, then `sort=<colum>&order=desc`, then no sort params, matching the arrows displayed, then I think this would give a positive indication to the user that the sort is by index (which is also the default).

Do you agree that's a suitable solution?

joe-crawford:
FYI while testing this I have found an issue with the backend not saving correct data element indexes and have raised as https://github.com/MauroDataMapper/mdm-core/issues/367, although I don't think it's related to this change.

GBishop-PHSA:
@joe-crawford I think fixing the cycle of asc/desc/no sort, would meet our requirements. Should I create a new ticket for the sorting cycle fix?

joe-crawford:
@GBishop-PHSA I don't think a new ticket would be required but might be good to update the existing gh-662. About the drag and drop to sort, I agree that's an issue, I'd suggest not allowing drag and drop unless the list is in index order (i.e. no arrows).

GBishop-PHSA:
@joe-crawford That's a great idea. I'll start working on the code change.

GBishop-PHSA:
@joe-crawford I've rolled back the original change and added the new rule to only allow drag/drop when the table is not sorted.  The test step is failing on jenkins after I rebased with the develop branch. I'm not sure how to re-run the build.

joe-crawford:
@GBishop-PHSA, this looks great, thanks.

joe-crawford (approved):

Issue #664, opened 2022-09-14
Classified Elements appear unordered and column headers do not order items


GBishop-PHSA:
# Description

In the Classifier Details screen, on the Classified Elements tab, the catalogue items are not ordered on screen entry and clicking the column headers does not result in the items being sorted.

# Steps to reproduce

1. Go to the Classifications tab
2. Select a classifier that has at least 2 catalogue items
3. Select the Classified Elements tab
4. The items are not sorted by Name
5. Click the Name or Description or Type headers
6. The sort ordering does not change

# Expected behavior

The sort order should change based on the column header that is clicked

# Environment

## Desktop 

- OS: Windows 10
- Browser: Chrome
- Version: 105.0.5195.102

# Additional context
The mdm-core controller does not read the sort query param and the UI is missing the domainType sort name

GBishop-PHSA:
Pull requests are ready for review:
https://github.com/MauroDataMapper/mdm-ui/pull/667
https://github.com/MauroDataMapper/mdm-core/pull/366

PR #667, opened 2022-09-28
gh-664 Add sort by catalogue item type

GBishop-PHSA:
gh-664 Fix for sort issue with Types column on Classifier Details > Elements tab


jamesrwelch (approved):
Good catch - thanks!

jamesrwelch:
closes #664

Issue #675, opened 2022-11-21
Show / Hide Editable after finalization alert messages using configuration


GBishop-PHSA:
# Description

When custom profiles have fields set to "May be edited after finalisation" and a user assigns the custom profile to a catalogue item, the application shows:
- alert message "This profile can be edited even after finalization" appears on the details screen
- alert message "Only certain fields can be edited on this profile as the catalogue item is finalized" appears on the edit screen

At PHSA, we set all fields in a profile to either editable after finalization or not-editable. Since a profile will not have a mix of editable/read-only fields, we won't need the alert messages for user. If the Edit button for a selected profile is visible, users will be able to edit any of the fields.

These alert messages could be hidden by setting a UI configuration in the Admin screen.

key:  ui.show_can_edit_property_alert
default value: true
category: UI

PR #676, opened 2022-11-22
gh-675 Show / Hide edit after finalization alerts

GBishop-PHSA:
gh-675 Added configuration that can hide the edit after finalization alerts found on profile edit and details screens.

config key: ui.show_can_edit_property_alert
default value: true
category: UI


jamesrwelch (approved):
This works really nicely.  It may be that in the future we change the styling, or allow alerts like this to be dismissed, never to be seen again, but this is a neat work-around for now.  Thanks for the submission!

Issue #371, opened 2021-10-20
The Ignore CTA on Link Suggestion functionality is not working and not refreshing the count online/real-time  


Jayafo:
**Describe the bug**
When a user uses the Link suggestion functionality trying to link a Data Asset, the **Ignore** CTA is not ignoring the data element and the count reader is just counting to infinity. Source Data Element DEC_4TH_FORENAME was clicked **250 times** whereas it only appears once on the list.  
**To Reproduce**
Steps to reproduce the behaviour:
1.	Go to 'Mauro'
2.	Select a model
3.	Select a sub folder
4.	Select the Data Asset in the view tree
5.	Select the Context menu
6.	Select Links in the dropdown menu list
7.	Select Link Suggestions
8.	A new page is open (Not sure if this is the process)
9.	Select a Target Data Model
10.	And select Suggest CTA
11.	Suggested links are populated 
12.	The **Ignore** CTA on the first Source Data Element (More than 150 times)
Expected behaviour
The Ignore CTA should remove the Source Data Element on the list and reflect on the count  

**Screenshots**
![image](https://user-images.githubusercontent.com/87768436/138130869-ef697ede-5f96-4898-8ec4-6fb41df41af9.png)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

Jayafo:
Passed Retest

PR #677, opened 2022-11-25
gh-371 Add filtering for classifier elements screen

GBishop-PHSA:
gh-371 Updated the classifier elements screen to show a filter text box for the Type column. Updated the element-link component with a new config property to show a domain type without the DataType prefix as users would use DataType as a filter value when filtering the results, and the elements' type value is not prefixed with DataType.

### DIFF
--- a/src/app/services/element-types.service.ts
+++ b/src/app/services/element-types.service.ts
@@ -153,8 +153,8 @@ export class ElementTypesService {
       title: 'DataType (Reference)',
       baseTitle: 'DataType',
       markdown: 'dt',
-      classifiable: true,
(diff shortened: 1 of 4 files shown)


joe-crawford:
I commented on the core PR about making the Type filter a dropdown (like the DataModel Data Types page filter) but have now noticed those dropdowns don't allow clearing properly so happy to leave that for now. I can raise a separate issue about that.

GBishop-PHSA:
@joe-crawford  This is ready for re-review

joe-crawford (approved):

Issue #375, opened 2021-10-22
Server error on sorting Data Types by Type


jamesrwelch:
On the data model page, when viewing the data types.  If you try and sort the data types by type, you get a server error.

As a side issue, the message displayed when there are no data types is slightly wrong - it should say "model" rather than "repository", and should point out that this may be because you've applied a filter.

PR #678, opened 2022-11-26
gh-375 Case insensitive Sorting for classifier elements

GBishop-PHSA:
gh-375 Set sorting to be case insensitive for classifier elements as catalogue items of different types may have different naming conventions - ie: all caps, sentence case or all lower case.


joe-crawford:
Should be unnecessary if updating https://github.com/MauroDataMapper/mdm-core/pull/376 to use a default of true.

GBishop-PHSA:
@joe-crawford  I'll close the pull request as it's no longer needed.

PR #692, opened 2022-11-28
gh-680 Make footer stick to bottom of page

### LINKED ISSUE #680: Sticky Footer
The Footer in the UI doesn't always stick to the bottom of the page - this is noticeable when on the search page, or when there are only a few items in the tree.  See attached screenshot.

The page structure should be updated so that the footer can be pushed to the bottom... this can be copied from #feature/style-tweaks

Screenshot 2022-11-28 at 10 02 37

pjmonks:
- When page contains additional space in content, footer will also attach to the bottom of the screen
- If the page height requires vertical scrolling, footer will always be at the end of the page
- Remove model tree forced height

Resolves #680 

image

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -22,13 +22,13 @@ SPDX-License-Identifier: Apache-2.0
 
     <main ui-view></main>
 
-    <mdm-footer
-      class="app-footer"
-      [version]="appVersion"
-      [copyright]="copyright"
-      [links]="footerLinks"
-    ></mdm-footer>
-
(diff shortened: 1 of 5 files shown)


jamesrwelch (approved):
Beautifully sticky, thank you!
I wonder whether the styling in the div in src/app/shared/models/models.component.html should be in a css file, but maybe there's another issue to be raised for moving all styling into css files rather than in the tags themselves.  
I'll accept this and enjoy the extra whitespace in the middle of the screen!

PR #698, opened 2022-11-29
gh-683 Add slider back between model tree and detail view

### LINKED ISSUE #683: Sliding two-column view
The two-column view that displays the tree and the model details could use a sliding component that allows users to adjust the width of the tree.  This will be necessary for #684 so that all the text can be read.  In the #feature/style-tweaks branch I've used https://angular-split.github.io and styled with a grey background for the body, with white panels, to more clearly see what's going on.  See screenshot

Screenshot 2022-11-28 at 10 30 26

pjmonks:
- Uses angular-split package
- Clean up unecessary code to handle resizing and responsiveness to layout

Resolves #683 

Desktop view:

image

Mobile view:

image

Collapsed model tree:

image


jamesrwelch (approved):
Lovely stuff - thanks!

Issue #687, opened 2022-11-28
Restyle tooltips


jamesrwelch:
Tooltips are currently quite small and hard to read.  Style them as per #feature/style-tweaks... See below for old / new
<img width="174" alt="Screenshot 2022-11-28 at 10 57 41" src="https://user-images.githubusercontent.com/1869668/204261522-8afb7a2c-e3ae-4ce4-b85b-39aab229fa7a.png">

<img width="147" alt="Screenshot 2022-11-28 at 11 00 21" src="https://user-images.githubusercontent.com/1869668/204261678-49415f2d-9f5d-40e2-9bc4-c5d383238e7d.png">

PR #699, opened 2022-11-29
Increase tooltip size to 12px

joe-crawford:
Partially resolves gh-687.

After change:
Screenshot 2022-11-29 at 17 32 45


pjmonks (approved):
I think this does improve the tooltip readability, it seems a lot clearer to me. Approved ✔️

Issue #690, opened 2022-11-28
Restyle tabs


jamesrwelch:
At the moment it's not always obvious that the tabs are tabs, and which is currently selected.  The changes in #feature/style-tweaks go some way to addressing this, although it may still be improved.
<img width="562" alt="Screenshot 2022-11-28 at 10 39 22" src="https://user-images.githubusercontent.com/1869668/204257526-d1d30d63-75f3-4ce8-8f4e-632590413c7e.png">

<img width="564" alt="Screenshot 2022-11-28 at 10 39 28" src="https://user-images.githubusercontent.com/1869668/204257511-44ce238f-cc14-476d-a8ca-41f750464817.png">

PR #700, opened 2022-11-29
Make active tab more prominent

joe-crawford:
Make sure opacity of active tab does not decrease when active but not focused, and make sure there is a difference in weight between the active tab and the rest (was an issue on NHS Digital theme).

Before:
Screenshot 2022-11-29 at 17 34 39
Screenshot 2022-11-29 at 17 39 25

After:
Screenshot 2022-11-29 at 17 35 09

Screenshot 2022-11-29 at 17 30 38

Partially resolves gh-690.


pjmonks (approved):
These changes look good 👍

PR #706, opened 2022-12-06
gh-639 Create consistent domain view header component

### LINKED ISSUE #639: Investigate combining detail view components
# Description

Currently each catalogue item type - such as Data Models, Terminologies, Folders, etc - are composed of two unique components:

- `[DomainType]Component` - the parent container to display all details and tabs
- `[DomainType]DetailComponent` - to display the core details of the catalogue item e.g. label, versions, menu actions etc

These 2 sets of components are duplicated for every domain type in Mauro, which means:

1. There can be a lot of duplicated code with only slight differences e.g. setting labels, finalising models etc
2. Adding new domain types takes a lot of copy/paste effort and testing
3. There are inconsistencies between all domain types. If one common action is added for all domain types, it produces a lot of UI effort to add to everything.

It should be investigated whether it is possible to create one detail view component that renders any catalogue item instead. There will need to be:

1. Some conditional logic to decide if certain UI controls/tabs are visible e.g. "Terms list" only appears for Terminologies
2. Most of the common logic would need to be pushed up into a service to decide the best endpoint to use e.g. setting a item label would use different endpoints depending on the domain type used.

# Initial Scope

For the initial investigation, it should be seen if one domain can be refactored into a more generic form such as:

1. The Data Model, since that should most operations you would expect. Or
2. Folders, since containers are quite simple catalogue items

Findings should then be discussed with the rest of team to determine next steps.

# Out of Scope

- Any redesigns or changes to UX. The existing functionality should remain in place, as well as how the pages look e.g. maintain the tab views
- Child components should be maintained and kept as they are for now. For instance, re-use the same Term/Data Element/Data Class list tables used already, depending on the domain type to represent

pjmonks:
- Replace all `*-detail` components which replicate each other with `mdm-model-header`
- All actions allowed for every domain type can now be done in here
- Changing label now launches a dialog instead of inline view
- Toolbar options changed around a bit to make common ones more prominent
- Every domain type view is now more consistent, removed all previous components to reduce code bloat

Resolves #639 

image

image


jamesrwelch:
@pjmonks My bad... it's ok, my laptop wasn't completely up to date.

jamesrwelch (commented):
I like the new 'edit title' dialog.  I think I'd pre-populate the text box with the existing title, but that can be another ticket

jamesrwelch (commented):
This goes further than I'd imagined - a box for all components, not just top-level models.  So the name of the new component is not quite right, but perhaps a different issue.

jamesrwelch (commented):
There's an issue where you're allowed to delete components of a finalised model, but I've checked and it was also an issue in the develop branch, so again, I'll create a new issue.

jamesrwelch (approved):
Lots of lovely code deletion!  This is a really good piece of work - I've tried to find problems and can't find any (other than the ones I've mentioned which were pre-existing).  Great work!

Issue #697, opened 2022-11-29
Classifier Elements grid shows superseded data models


GBishop-PHSA:
# Description

The classifiers details screen > Element tab searches for all catalogue items with the associated classifier. Users may not want to see the superseded data models, as they are also hidden on the browser tree. 

The API will return results for a classifier without supersended documents if the query parameters: includeDocumentSuperseded 
and includeModelSuperseded, are set to false.

GBishop-PHSA:
This required changes to mdm-ui and mdm-core
Pull requests: 
https://github.com/MauroDataMapper/mdm-core/pull/380
https://github.com/MauroDataMapper/mdm-ui/pull/707

PR #707, opened 2022-12-06
gh-697 Hide superceded documents in Classifier Elements list

GBishop-PHSA:
gh-697 Pass the parameters to not show superceded documents on the classifier elements tab screen


joe-crawford:
I think the linter is complaining about missing semicolons, could you update that?

We're having an issue with the network connection from to GitHub on our end, I'll merge when that's working again.

joe-crawford:
Hi @GBishop-PHSA, closing this PR due to the linting issue and because I think the defaults are already applied through the mdm-core PR, please reopen if you want to include this.

PR #712, opened 2022-12-08
gh-709 Improvements to the Rules and Representations UI

### LINKED ISSUE #709: Improve rules and rule representations
Make improvements to the "Rules" and rule representations UI. This is to support work required for the Mauro Data Explorer.

- Make sure there are proper type definitions on the endpoints (`mdm-resources`)
- Make sure the UI code is correct
- Add in new rule representation languages required for the Cohort Builder - see MauroDataMapper/mdm-explorer#164.

pjmonks:
*Note:* requires MauroDataMapper/mdm-resources#79 to be merged first.

Resolves #709

- Use type definitions for rules from mdm-resources
- Fix sort order of rule languages filter
- Code cleanup
- Improve Add Rule  and Add Rule Representation dialog - use reactive forms, add type definitions
- Add JSON support to rule representations
- Hide export option if rule has no representations
- Improve expand/collapse buttons for representations
- Move common action buttons out of context menu and into the table row
- List the number of representations in the rules table, then shows if there are any

image

image

### DIFF
--- a/src/app/constraints-rules/constraints-rules.component.html
+++ b/src/app/constraints-rules/constraints-rules.component.html
@@ -170,12 +170,12 @@ SPDX-License-Identifier: Apache-2.0
               <span
                 *ngIf="(record.ruleRepresentations?.length ?? 0) > 0"
                 [ngClass]="
-                  this.expandedElement === record
+                  this.expandedRuleId === record.id
                     ? 'fas fa-angle-double-up'
                     : 'fas fa-angle-double-down'
                 "
                 [matTooltip]="
-                  this.expandedElement === record ? 'Collapse' : 'Expand'
+                  this.expandedRuleId === record.id ? 'Collapse' : 'Expand'
                 "
               ></span>
             </button>
@@ -266,7 +266,7 @@ SPDX-License-Identifier: Apache-2.0
             <div
               class="example-element-detail"
               [@detailExpand]="
-                row == expandedElement ? 'expanded' : 'collapsed'
+                row.id == expandedRuleId ? 'expanded' : 'collapsed'
               "
             >
               <table
--- a/src/app/constraints-rules/constraints-rules.component.html
+++ b/src/app/constraints-rules/constraints-rules.component.html
@@ -24,59 +24,66 @@ SPDX-License-Identifier: Apache-2.0
         [hideToggle]="true"
         [expanded]="filteredOpen"
         class="prevent-click mat-elevation-z0"
-        (click)="clickButton=false"
+        (click)="clickButton = false"
       >
         <mat-expansion-panel-header>
+          <div
+            fxFlex
+            fxLayout="row"
+            fxLayout.md="row"
+            fxLayout.sm="row"
+            fxLayout.xs="column"
+            fxLayoutAlign="space-around"
+          >
             <div
               fxFlex
               fxLayout="row"
-              fxLayout.md="row"
-              fxLayout.sm="row"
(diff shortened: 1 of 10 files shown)


jamesrwelch:
https://github.com/MauroDataMapper/mdm-resources/pull/79 doesn't allow the develop branch of the UI to be run up, so I propose that when we're happy with this branch, we mark this one as 'ready' and do them both at the same time.

jamesrwelch:
I wonder whether we should allow different types of JSON representation?  E.g. have `JSON (MEQL)` as an option for now, and allow other types of JSON later?

jamesrwelch:
I'm happy with all the improvements here - well done :)

pjmonks:
> I wonder whether we should allow different types of JSON representation? E.g. have `JSON (MEQL)` as an option for now, and allow other types of JSON later?

You could potentially end up with a lot of "JSON" formats, since JSON is quite popular as a representation. There could be further changes, I'm still not sure how many representations will be stored in Mauro for a cohort query - just JSON or the other formats like SQL or MEQL?

NigelPalmer:
"I'm still not sure how many representations will be stored in Mauro for a cohort query ", my view is that there should be 1 representation to rule them all. The other representations should be transformed from that representation. If we start saving multiple copies of the same data we run the risk of them going out of sync with each other. If things get out of sync then maintaining the data becomes a burden and confidence in data quality is damaged.

jamesrwelch:
> my view is that there should be 1 representation to rule them all. 
Agreed.

I'd suggest renaming `JSON` to `JSON (MEQL)` (or `MEQL (JSON)`) and we can add other JSONs as and when we find them.

NigelPalmer (changes requested):
**Passed**

- [x] JSON (MEQL) has been added. 
- [x] Rules save successfully.
- [x] Representations save successfully.
- [x] Representations are editable. 

 

**Issues**
It is still possible to get into a state where the buttons are inaccessible. The most likely case where this would be an issue would be someone running their browser full screen at 200% zoom.
image

Could we try and do something so that it is possible to scroll the buttons into view if they are off the bottom of the screen?

NigelPalmer:
Vertical scroll bar now present when buttons are off the bottom of the screen. Much better.

@pjmonks , could you please respond to James thoughts in some way. They could always be addressed in a follow-up issue if it makes sense to do so.
> I'm happy with all the improvements here - well done :) Some thoughts for discussion:
> 
> * When you add a new representation, the list of representations for that rule gets hidden - are we able to save that state?
> * The display of rules / representations is quite verbose - is there anything we can do to improve it?
> * The button that exports all rules works in a different way to the normal export data model functionality - are we ok with this?
> * Really long representations get cut off - could we maybe use the more/less component from descriptions?

pjmonks:
@NigelPalmer I've updated two minor things - keep rules open on adding new items and add a more/less area - and separated other points into other issues.

NigelPalmer:
- [x] Rules stay open when adding new representations.
- [x] More link available 
- [ ] Less link not displayed when expanded. It is not possible to collapse a section. **Please fix.**
- [x] Follow-up tasks created to address all other issues raised by James.

pjmonks:
@NigelPalmer created a separate issue, #716, to counter the last issue raised. If happy with everything else, please approve the review so this can be merged.

NigelPalmer:
- [X] Follow up task created to address missing "Show Less" option.
This is good to go now.

NigelPalmer (approved):

PR #713, opened 2022-12-09
gh-710 Prepopulate Change Label dialog with original label

### LINKED ISSUE #710: When editing a label, show previous value
The new pop-up for editing the label of an item is great, but it would be convenient if the text box was pre-populated with the existing label (for when minor changes are required).
Screenshot 2022-12-07 at 17 10 57

pjmonks:
Useful for users to start from something when making changes

Resolves #710


jamesrwelch (approved):
Lovely stuff, thank you!

PR #725, opened 2023-01-19
Update scss imports to not use '~'

### LINKED ISSUE #724: Update SCSS imports to match latest recommendations
The '~' in import statements in SCSS files is now deprecated - by default the `node_modules` folder will be searched if the file is not found within the app.  See https://webpack.js.org/loaders/sass-loader/#resolving-import-at-rules for details.  
This will cause issues during the upgrade to Angular 13 / 14, so presented as a separate task here for easier review.

jamesrwelch:
Fixes #724.  For merging before upgrade to angular 14


pjmonks (approved):
Still compiles and looks the same. Approved ✔️

PR #727, opened 2023-01-19
Feature/gh 726

### LINKED ISSUE #726: Bring in Idle Service
We currently depend on the User Idle service here: https://github.com/rednez/angular-user-idle
This doesn't yet run on Angular 14, and will cause us dependency conflicts when we try to upgrade our dependencies.  Since the package only provides one service, and is MIT licensed, let's bring the files into our codebase for now, and reconsider in the future.

jamesrwelch:
Fixes #726 In preparation for the Angular 14 upgrade

### DIFF
--- a/license-check-and-add-config.json
+++ b/license-check-and-add-config.json
@@ -27,7 +27,8 @@
     "**/*.ttf",
     "**/*.woff",
     "**/*.woff2",
-    "sonar-project.properties"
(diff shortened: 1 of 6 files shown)


pjmonks (approved):
Looks as I would expect based on same work for `mdm-explorer`.

Approved ✔️

PR #730, opened 2023-01-19
Properly type, and make consistent, all `unsubscribe$` `Subject` definitions

### LINKED ISSUE #729: Properly type unsubscribe Subjects
A lot of components declare an `unsubscribe$` Subject and specify it without a dynamic type:

```  
private unsubscribe$ = new Subject();
```

These should be defined using a dynamic type to help the type checker:

```
private unsubscribe$ = new Subject<void>();
```
This is not an issue at the moment, but after an upgrade to angular 14, and the latest typescript libraries, it will cause issues.

jamesrwelch:
Fixes #729 and helps simplify the upgrade to angular 14.


pjmonks (approved):
Good additions for type checking 👍

PR #732, opened 2023-01-19
Rename outputs to match future lint rules

### LINKED ISSUE #731: Future lint issue: 'on' prefix for event names
According to the angular style guide, event names shouldn't begin with 'on': https://angular.io/guide/styleguide#style-05-16
When we upgrade to angular 14, this will become a blocking issue, so setting it up to be solved before we get that far.

jamesrwelch:
Fixes #731 - renaming of output / events to fix future lint issues in Angular 14

### DIFF
--- a/src/app/bulk-edit/bulk-edit-select/bulk-edit-select.component.ts
+++ b/src/app/bulk-edit/bulk-edit-select/bulk-edit-select.component.ts
@@ -61,8 +61,8 @@ interface CatalogueItemDomainTypeOption {
   styleUrls: ['./bulk-edit-select.component.scss']
 })
 export class BulkEditSelectComponent implements OnInit, OnDestroy {
-  @Output() onCancel = new EventEmitter<void>();
-  @Output() onNext = new EventEmitter<void>();
+  @Output() cancel = new EventEmitter<void>();
+  @Output() next = new EventEmitter<void>();
 
   /** Two way binding */
   @Input() context: BulkEditContext;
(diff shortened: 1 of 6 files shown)


src/app/bulk-edit/bulk-edit-select/bulk-edit-select.component.ts
```diff
@@ -61,8 +61,8 @@ interface CatalogueItemDomainTypeOption {
   styleUrls: ['./bulk-edit-select.component.scss']
 })
 export class BulkEditSelectComponent implements OnInit, OnDestroy {
-  @Output() onCancel = new EventEmitter<void>();
-  @Output() onNext = new EventEmitter<void>();
+  @Output() cancel = new EventEmitter<void>();
+  @Output() next = new EventEmitter<void>();
```
pjmonks, line 65: The output bindings have been changed in these child controls, but the parent components that use these have not. Currently, when testing the bulk editor, I cannot progress forward or backward after making selections. The `mdm-bulk-edit-container` component also needs to be updated:

```html
<!-- Current usage of the components, using wrong output names -->
<mdm-bulk-edit-select
  *ngIf="currentStep === Steps.Selection"
  [(context)]="context"
  (onNext)="next()"
  (onCancel)="cancel()"
></mdm-bulk-edit-select>
```

pjmonks (changes requested):
Testing has shown that changing the output binding names may have broken some functionality, since parent component templates have not been updated to match new names - though I would have thought some compilation errors would have occurred.

Please check any components that reference these and update their bindings if required:

- `mdm-bulk-edit-editor-group`
- `mdm-bulk-edit-select`
- `mdm-element-label`

src/app/bulk-edit/bulk-edit-select/bulk-edit-select.component.ts
jamesrwelch: Thank you - good catch.  It appears to be just that one instance that has an issue.  It has a binding for  `[(context)]` that doesn't appear to exist.  Shall I ditch it now, or leave it for another tidying session?

PR #734, opened 2023-01-19
Update Angular Material import locations

### LINKED ISSUE #733: Update angular imports
There are a few modules imported from angular as follows:

```
import { ComponentRef } from '@angular/core/core';
```

These should be as follows:

```
import { ComponentRef } from '@angular/core';
```
And they'll break when we update to Angular 14.

jamesrwelch:
Fixes #733 and simplifies the upgrade to angular 14

### DIFF
--- a/src/app/utility/extensions/mat-dialog.extensions.ts
+++ b/src/app/utility/extensions/mat-dialog.extensions.ts
@@ -57,7 +57,7 @@ import {
 import { Observable } from 'rxjs/internal/Observable';
 import { filter, map, mergeMap } from 'rxjs/operators';
 
-declare module '@angular/material/dialog/dialog' {
+declare module '@angular/material/dialog' {
   interface MatDialog {
     /**
      * Extension method to open a modal dialog containing the `ConfirmationModalComponent`.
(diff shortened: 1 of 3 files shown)


src/app/utility/extensions/mat-dialog.extensions.ts
```diff
@@ -57,7 +57,7 @@ import {
 import { Observable } from 'rxjs/internal/Observable';
 import { filter, map, mergeMap } from 'rxjs/operators';
 
-declare module '@angular/material/dialog/dialog' {
+declare module '@angular/material/dialog' {
```
pjmonks, line 60: This change produces a compilation error (at least in the current version of `mdm-ui` using Angular 12). This is because the `declare module` is trying to reference the exact JavaScript file where the `MatDialog` type is located, so `@angular/material/dialog/dialog` is correct - equates to the file `node_modules/@angular/material/dialog/dialog.d.ts`. Changing the module reference means all the extension methods cannot be bound.

pjmonks (changes requested):
Compilation errors I'm afraid ❌ See my comment for details.

src/app/utility/extensions/mat-dialog.extensions.ts
jamesrwelch: @pjmonks Good point - I've made that change in my bigger branch successfully, but maybe it needs making once angular has been updated.  I've reverted this one for now and hopefully it'll go through this time

pjmonks (approved):
Compiles now ✔️

PR #743, opened 2023-01-27
gh-739 Fix Bulk Editor cancel action

### LINKED ISSUE #739: [Post Angular 14 upgrade] Cannot cancel and go back in Bulk Editor
Bug raised as part of testing #738 

# Description

You are now unable to go back/cancel from inside the Bulk Editor of a model. Also, no confirmation dialog appears for the user to continue or cancel the operation.

# Steps to reproduce

Assuming you are testing on the version of `mdm-ui` which was upgraded to Angular 14 - see PR #738:

1. Go to any model in the tree in draft mode
2. In the three dot menu, select "Bulk Edit"
3. Click on the "Cancel" button, or
4. Select items/options to progress to the grid editing page, then click the "Close" button.

# Expected behavior

Clicking "Cancel" or "Close" should display a confirmation dialog. If the user clicks "Yes", then go back to the previous page. If the user clicks "No", remain in the Bulk Editor.

# Additional context

The confirmation dialog should appear automatically as part of the `UIRouter` transition hooks for leaving a page:

1. The Bulk Editor pages sets `EditingService.start()`
2. The cancel/close operation should make the UI router state handler go back
3. The transition hook then triggers the `EditingService.confirmStopAsync()` function which should display the `ConfirmationModalComponent` inside a `MatDialog`.

All of this happens, but for some reason the dialog triggers a close immediately. This only seems to happen in the Bulk Editor, other locations in the app don't have this problem.

The workaround to fix this issue is to manually display the confirmation dialog and not rely on the transition hook, by modifying the `bulk-editor-container` component:

```ts
cancel() {
  this.editing.confirmCancelAsync()
      .pipe(filter(confirm => !!confirm))
      .subscribe(() => {
        this.editing.stop();
        this.stateHandler.GoPrevious();
      })
}
```

pjmonks:
Manually confirm with user they want to cancel and go back a page if they confirm.

Fixes #739

### DIFF
--- a/src/app/bulk-edit/bulk-edit-container/bulk-edit-container.component.ts
+++ b/src/app/bulk-edit/bulk-edit-container/bulk-edit-container.component.ts
@@ -24,7 +24,7 @@ import { StateHandlerService, MessageHandlerService } from '@mdm/services';
 import { EditingService } from '@mdm/services/editing.service';
 import { UIRouterGlobals } from '@uirouter/core';
 import { EMPTY } from 'rxjs';
-import { catchError } from 'rxjs/operators';
+import { catchError, filter } from 'rxjs/operators';
 import { BulkEditContext, BulkEditStep } from '../bulk-edit.types';
 
 @Component({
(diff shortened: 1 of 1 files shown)


jamesrwelch (approved):
Lovely stuff - it was a good catch in the first place.  The code looks good; I've had a play and can't break it.  10/10 - would buy again.

PR #745, opened 2023-01-30
gh-742 Code updates to support Angular 14 typed forms

### LINKED ISSUE #742: Use Angular 14 Typed Forms
Angular 14 introduces Typed Forms:  https://angular.io/guide/typed-forms
If we were to use these in our code we'd get better type-checking and hopefully better code quality.
The current code uses the default `UntypedForm` in the few places where Forms are actually defined - these need converting to proper Typed versions, and there are probably other places in the code where explicit Form definitions could be introduced.

pjmonks:
Resolves #742 

Note that some existing forms have _not_ been updated as they seem more complex, these are:

- `mdm-new-version`
- `mdm-openid-connect-provider`
- `mdm-create-term-relationship-dialog`

All other uses of `FormGroup` have been updated to be strongly typed.

### DIFF
--- a/src/app/admin/api-property/api-property.component.ts
+++ b/src/app/admin/api-property/api-property.component.ts
@@ -20,10 +20,23 @@ import { Component, OnInit } from '@angular/core';
 import { FormControl, FormGroup, Validators } from '@angular/forms';
 import { MatSelectChange } from '@angular/material/select';
 import { Title } from '@angular/platform-browser';
-import { ApiProperty, ApiPropertyIndexResponse, ApiPropertyResponse } from '@maurodatamapper/mdm-resources';
-import { ApiPropertyEditableState, ApiPropertyEditType, ApiPropertyMetadata, propertyMetadata } from '@mdm/model/api-properties';
+import {
+  ApiProperty,
+  ApiPropertyIndexResponse,
+  ApiPropertyResponse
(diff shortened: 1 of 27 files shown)


jamesrwelch:
@pjmonks The conflicts on this branch look fairly straightforward, but could you resolve them so I don't make any horrible mistakes?  Thanks!

jamesrwelch (changes requested):
This is really great work - this will help make our code much more reliable.  I've played with all the forms affected by this change, and unfortunately I've found one that no-longer works - the one for adding / editing a subscribed catalogue.  If you can fix that one, then I think this is good to go.  Thanks!

jamesrwelch (changes requested):
Should there be more to this change?  The max / min in the control work, but I can't get it to show the red error messages?

pjmonks:
> Should there be more to this change? The max / min in the control work, but I can't get it to show the red error messages?

There are two ways to enter a number. Either click on the up/down arrows of the numeric control - then the min/max constraints will not allow you to go above/below those values.

Alternatively, type in a number manually in the text input. Then leave focus of that field and the validation should kick in, forcing you to enter a value within those constraints.

jamesrwelch:
You're right - it's the changing focus bit that I'd missed.   I'm happy with this.

jamesrwelch (approved):

PR #747, opened 2023-02-01
gh-422 Fix category selection in Enumeration Value table

### LINKED ISSUE #422: Enumeration Data Type - Grouping
**Describe the bug**
Entering grouping in the enumeration types seem to allow only one character, and user is not sure whether the value is entered unless user saves the new entry.

**To Reproduce**
1. Create an enumeration type, and Add enumeration
2. Enter Key / Value pair
3. Enter chars into group input box

As soon as user enters one character, the cursor moves out of the group box. After clicking on the Sasve button, the key/value/group is saved, but group has only one character.

**Expected behavior**
User shall be able to name a group. The cursor should not escape the input box.

**MDM version**
MDM 4.10/6.6.0

**Screenshots**
image

pjmonks:
- Replace mdm-select component with mat-autocomplete. More reliable when selecting options
- Allows the ability to enter new values through text when options for categories are not present
- Now correctly saves category names to enumeration values
- Fix drag and drop interfering with row edits

Resolves #422 

**Note:** I've noticed other issues with this table control (it does feel like the most dependable) but this PR only focuses on the issue raised in #422 regarding saving category names.

image


jamesrwelch (approved):
This works a whole lot better now, thank you so much!  If there are remaining issues, could you create tickets for them before we forget?

Issue #88, opened 2021-05-10
Ensure consistency with "Log in" and "Log out" buttons


pjmonks:
**Describe the bug**
[End to end tests](https://github.com/MauroDataMapper/mdm-end-to-end-testing) can fail on development instance because of inconsistency between buttons for logging in and out.

**To Reproduce**
1. Setup UI to run as: `npm run start -- --post 8082`
2. In the `mdm-end-to-end-testing` repo:
    1. Run `npm run webdriver-start` to start the Selenium web driver
    2. Run in another process `npm run build` then `npm run test` to start the e2e tests.

**Expected behavior**
All tests should pass with a fully green report.

**Additional context**
The issue is because the test steps are searching for a "Log in" button on the modal form to submit, when the actual text is "Login".

This should be fixed in the UI as follows:

1. Ensure all buttons for logging in are labelled "Log in", matching casing and the included space.
2. Ensure all buttons for logging out are labelled "Log out", matching casing and the included space.

PR #753, opened 2023-02-06
Update to node version 16.10.0

### LINKED ISSUE #748: Upgrade to Node 16 with updated npm
# Description

Consider upgrading NodeJS to the next LTS version, which should also include an upgrade to npm.

# Additional context

Currently, the repo targets Node 14.18.1 (as specified by the `.nvmrc` file). This is still acceptable for Angular 14, but may become an issue for further upgrades in the future. According to https://unpkg.com/browse/@angular/core@14.2.12/package.json, Angular 14 supports:

```
"engines": {
    "node": "^14.15.0 || >=16.10.0"
  },
```

Suggestion is to upgrade to NodeJS 16.x.x at least.

Another reason to upgrade is that warnings can appear when running `npm install` regarding `package-lock.json` files being different versions. Ensuring the npm is in line with a later node version, this warning should then disappear.

# Actions

- [ ] Update `.nvmrc` file
- [ ] Install Node 16.x.x and update npm
- [ ] Test to ensure everything still builds and works

jamesrwelch:
Fixes #748 
Depends on https://github.com/MauroDataMapper/mdm-resources/pull/88


pjmonks (changes requested):
There are build issues here. Jenkins reports this error when compiling and running tests:

```
Cannot find module 'tslib' from '../../.nvm/versions/node/v16.10.0/lib/node_modules/@maurodatamapper/mdm-resources/lib/es2015/index.js'
```

I also had this problem building locally but now have gotten into a state where I can build it locally but not on the server. I think the issue is isolated to this branch.

If you remove `importHelpers` from `tsconfig.json` that seems to resolve the issue, but we're not sure if that is something that should be disabled. The setting also seems linked to `downlevelIteration` - again, something worth checking if we need, do we need to "downlevel" JavaScript code if the `target` property is set to `es2020`?

Another thing I attempted was to upgrade the `tslib` package in `package.json` but this seems to have no affect.

Please review this issue and try again.

jamesrwelch:
@pjmonks Some different failures now, which I'm looking into.  I couldn't reproduce the build issues locally, once I'd got rid of them the first time.  Fiddling on the build server, the only change I made between it not working, and then working, was to manually re-link `mdm-resources`.  Not even close to understanding why...

jamesrwelch:
(I was led down that road because mid-way through the warnings were a couple about features of `mdm-resources` that didn't exist except in the latest version)

jamesrwelch:
Another build, with no other changes, and now it builds.  Going to fiddle a bit more to see if I can understand anything at all

jamesrwelch:
This is now fixed, basically by re-building and publishing mdm-resources.  I can't get it to fail again locally, or on the Jenkins server

pjmonks (approved):
I can't find any build issues now so let's assume it's all working normally now 🤞 

I've been using Node 16.10.0 locally without issue for a while once my initial build errors passed and it seems fine, ready to merge ✔️

PR #757, opened 2023-02-10
gh-756 Update bulk editor to differentiate between data type domains

### LINKED ISSUE #756: Bulk editing of data types does not return all useful profiles
# Description

If bulk editing a data model and choosing to edit data types, the list of available profiles may not include everything that is useful.

# Additional context

Depending on the underlying domain type of the data type - primitive, enumeration, model etc - this currently determines the profiles endpoint to use. For example, if only `/api/enumerationTypes/{id}/profiles/unused` is called, then other profiles for other domain types will not be available.

pjmonks:
Data types are actually defined further by being Primitive, Enumerations or Model. Profile endpoints need the precise domain type to fetch the correct lists of profiles available.

image

Resolves #756 

A good way to test would be on the "modules" data model as that should contain an appropriate mixture of data type domains.


jamesrwelch (approved):
Works really nicely, thank you!

PR #758, opened 2023-02-13
gh-755 Refactor group-access-new component

### LINKED ISSUE #755: Group roles do not appear in 'User & Group Access' modal
# Description

Group roles do not appear in the 'User & Group Access' modal, so user groups cannot be assigned new group roles. Screenshot below:
Screenshot 2023-02-10 at 10 56 16

# Steps to reproduce

In a Catalogue Item, click on the 'User & Group Access' button, then '+ Add Group', then select a Group name, then try to select an Access Level (group role), no Access Levels appear.
# Expected behavior

The available Access Levels (group roles) for the resource should appear in the dropdown.

# Environment

## Desktop 

- Browser: Chrome

## Mauro

- `develop` of core, ui and resources

pjmonks:
**Note:** depends on MauroDataMapper/mdm-resources#96 first

- Use stricter typing on endpoints and lists used
- Rename properties to make more sense
- Rename table columns
- Update add/delete options for managing securable resource groups
- Fix issue where selecting options when adding new groups does not work, replace custom controls with MatSelect

Fixes #755


joe-crawford (approved):
Great, thanks for this @pjmonks, this looks good to me and works well locally :+1:

PR #762, opened 2023-02-16
Add security.hide.exception API property

joe-crawford:
Add `security.hide.exception` API property to the UI list of configurable properties.

Depends on https://github.com/MauroDataMapper/mdm-core/pull/396.


pjmonks (changes requested):
I've tested this with the mdm-core branch it relates to and it appears to work when I've forced a backend server error to occur, the stack trace is not included when the API property is enabled.

One improvement I think I would like it to make an adjustment to this screen:

image

When there is no exception in the response it still adds an "Exception" row with an undefined value. If you look at `server-error.component.ts` you should be able to exclude that row if the exception is not present which will present a lot better.

joe-crawford:
Thanks @pjmonks, I've updated to hide any blank rows in the Server Error component.

pjmonks (approved):
I've re-tested and I'm happy with how the error screen looks now, approved ✔️

PR #763, opened 2023-02-20
Fix import stopping dist build on Jenkins

jamesrwelch:
This should fix the issue currently appearing in Jenkins:
```
Error: src/app/admin/subscribed-catalogue/subscribed-catalogue.component.ts:37:27 - error TS2307: Cannot find module '../../../../../mdm-resources/src' or its corresponding type declarations.
37 import {MdmResponse} from '../../../../../mdm-resources/src';
```


joe-crawford (approved):
Thanks for fixing that!

PR #764, opened 2023-02-20
Fix footer image path

jamesrwelch:
This fixes the pathing issue that's stopping the image appearing in the footer on our hosted instances


pjmonks (approved):
Works better now 👍

PR #770, opened 2023-02-27
gh-760: adding colour display for theme properties

### LINKED ISSUE #760: Theming: Display chosen colour next to Api Property value
# Description

In the future we may add a separate page for theming, but for now we are simply using the Api Properties page to theme the Mdm Explorer. A quick win to improve the UX is to display the chosen colour next to the hex value. The chosen colour should also be displayed on the edit api property page after the user has type in a hex value.

# Additional context

The following is a mockup of how I envision the api properties page:
image

The name of the key should be read to determine whether the colour box should be displayed or not. I.e if the name of the key starts with "explorer.theme.color" then display the coloured square.

The following is a mockup of how I envision the epi property edit page:
image

# Out of Scope
The user **will not** have any GUI for picking colours. The coloured box is simply to represent the typed in value.

CristianGimenez:
On property display, added a disabled input of type color to show the selected colour of the property value.
On property edit/add if the key of the property contains "theme.color" show 2 inputs, the user can use both of them, one is an input text where the hex code of the colour can be pasted, the other is an input of type colour so we can have an html colour picker.

As discussed in a conversation with @NigelPalmer,  in the issue description we say to check the key of the property and look for those which starts with "explorer.theme.color" but after some thought we changed that to be properties that contain "theme.color" since in the future this might be used to also theme mdm-ui.

Closes #760

### DIFF
--- a/src/app/admin/api-property-table/api-property-table.component.html
+++ b/src/app/admin/api-property-table/api-property-table.component.html
@@ -63,11 +63,11 @@ SPDX-License-Identifier: Apache-2.0
                 <span>Value</span>
             </th>
             <td mat-cell *matCellDef="let record" style="text-align: left; word-wrap: break-word;">
-                <span *ngIf="record.metadata.key.includes('theme.color')">{{record.original?.value}} <input  type="color" value="{{record.original?.value}}" disabled class="mdm-api-property-table__colour-display"></span>
-                <span *ngIf="!record.metadata.key.includes('theme.color') && (record.metadata.editType === EditType.Text || record.metadata.editType === EditType.Html)" class="mdm-api-property-table__value--content">
+                <span *ngIf="record.metadata.key | matchThemeColorPattern">{{record.original?.value}} <input  type="color" value="{{record.original?.value}}" disabled class="mdm-api-property-table__colour-display"></span>
+                <span *ngIf="!(record.metadata.key | matchThemeColorPattern) && (record.metadata.editType === EditType.Text || record.metadata.editType === EditType.Html)" class="mdm-api-property-table__value--content">
                     Content
                 </span>
-                <span *ngIf=" !record.metadata.key.includes('theme.color') && (record.metadata.editType === EditType.Value || record.metadata.editType === EditType.Boolean)">{{record.original?.value}}</span>
+                <span *ngIf=" !(record.metadata.key | matchThemeColorPattern) && (record.metadata.editType === EditType.Value || record.metadata.editType === EditType.Boolean)">{{record.original?.value}}</span>
             </td>
         </ng-container>
           <ng-container matColumnDef="icons">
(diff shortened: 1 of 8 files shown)


pjmonks (changes requested):
Having tested this it seems to work well. And although a colour picker control was originally out of scope, the fact that one is present in the add/edit property screen is very useful.

There is one big issue linked to the Mauro Data Explorer theme though, in that the current API keys used for the MDE theme do not match the expression "theme.color", so all current keys listed for the MDE do not show these colour options. Issue #760 wasn't updated with the correct task description I think - look at MauroDataMapper-Plugins/mdm-plugin-explorer#18 for the correct table of possible values.

This means some of this logic will have to change accordingly to match.

CristianGimenez:
#Thanks @pjmonks, you were right, the issue wasn't up to date. After reviewing the list of properties added in [MauroDataMapper-Plugins/mdm-plugin-explorer#18](https://github.com/MauroDataMapper-Plugins/mdm-plugin-explorer/issues/18)... the pattern I've seen is that all of the colour properties there, have a part of the key that includes "theme" and later another part that includes "color", therefore, I've changed the using of `String.includes` to a regular expression match. Any key that contains:

- any or no characters (not numeric), followed by
- the string "theme" followed by,
- any or no characters (not numeric), followed by
- the string "color", followed by,
- any or no characters (not numeric)

will pass the regex expression which is` \D*theme\D*color\D*` by the way.

As discussed, I didn't see how to avoid having a call to a function in the template, which causes the function to be called each time Angular checks for changes (which can be a lot). Reading around I've found [this article ](https://medium.com/showpad-engineering/why-you-should-never-use-function-calls-in-angular-template-expressions-e1a50f9c0496#:~:text=The%20solution%3A%20how%20to%20avoid%20unnecessary%20function%20calls&text=A%20first%20strategy%20to%20reduce,pipe's%20input%20does%20not%20change.) that states pure pipes as an alternative. For the tests I've made it seems not to be called in a regular basis. So I have:
- created a pipe
- added tests for that pipe to check that the regular expression works as expected
- added the pipe to the admin module
- substituted the `String.includes` calls with the newly created pipe

I think this is ready to review again. Thanks for your help!

pjmonks (dismissed):
This looks better now. As far as I can tell, the table is not being constantly rendered thanks to the pipe used, which gets around having logic be part of the data binding 👍 

image

Approved ✔️

jamesrwelch:
Could I make one minor styling request?  Could the colour square be horizontally aligned on the right of the text?  Or on the left of the text as I suspect that might be easier?  Where you have short colours (`#fff`), and long colours, (`#000000`), the squares wiggle all over the place :)

CristianGimenez:
Updated after the comments from @jamesrwelch, now the span with the colour text or the input text for the edit page, have a fixed width of 100px, that will make the coloured square to be in place even with smaller text lengths

pjmonks (dismissed):
Colour values in table now look aligned:

image

Approved ✔️

pjmonks (approved):

jamesrwelch:
Ooh, very aligned, I like it 👍 
Not really a discussion for this PR, but perhaps a wider one - `px` vs. `em`/`rem`?

pjmonks:
@jamesrwelch Could you merge this please? Everything is approved but it's waiting for the license/cla check for some reason.

PR #772, opened 2023-02-28
gh-279 Remove restriction for signed in users to include superseded models

### LINKED ISSUE #279: Allow public users to view superseded models
The 'show superseded models' is currently in the user preferences.  It would be nice if there was also the same option on the tree view - in particular when a user is not logged in, so that public users can view superseded models.

pjmonks:
Anonymous users can now access this filter in the model tree and maintain their preferences locally.

Resolves #279 

image


jamesrwelch (approved):
Very happy with this - thanks!

PR #776, opened 2023-03-02
gh-648 Update loading indicator to always close when requests finish

### LINKED ISSUE #648: "Working" notification not correctly clearing on filtering
When you filter in a table, a request is made and the "Working" notification appears until the response returns.  If you keep typing in the filter field, the first request is cancelled, and a second one created - this is as expected.  However, when that second response returns, the "Working" notification doesn't get cleared, leaving the user the impression that something is still happening.

pjmonks:
Updated `HttpRequestProgressInterceptor` to handle HTTP requests that may be aborted or cancelled, these were being missed so left the "Working..." indicator visible.

Fixes #648 

# Testing

You can test this by finding a large data model and finding a table that allows filtering e.g. data types, data classes, data elements etc. Toggle the filter and type fast in a filter field - this should cause multiple HTTP requests to be sent and, if fast enough, some earlier ones may be cancelled/aborted. Even so, the "Working..." indicator should now always disappear eventually.

image

### DIFF
--- a/src/app/services/http-request-progress.interceptor.ts
+++ b/src/app/services/http-request-progress.interceptor.ts
@@ -16,10 +16,15 @@ limitations under the License.
 SPDX-License-Identifier: Apache-2.0
 */
 
-import { HttpEvent, HttpHandler, HttpInterceptor, HttpRequest, HttpResponse } from '@angular/common/http';
+import {
+  HttpEvent,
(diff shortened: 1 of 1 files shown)


jamesrwelch (approved):
This works really nicely and makes a big difference to user experience - thanks!

Issue #779, opened 2023-03-03
Pass profile filters from Search page to Search Listing page to activate search


pjmonks:
# Description

Once #777 is complete, the selected profile filters must be passed to the Search Listing page to trigger the search.

# Background

All parameters for the search from the Search page are passed as query parameters to the Search Listing page. The query parameters are used to:

1. Pass the correct information to the page and start the search
2. Allow the search to be repeatable

Now that profile fields are being added and can be variable in length, this poses more challenges since the URL query string to grow to be too long.

# Implementation

Below is one suggestion for maintaining the URL requirements above and also including the necessary parameters. It involves passing the profile filters as a JSON object via the URL.

## Structure the profile fields

Collect the profile filters chosen from the UI into a JSON object. This should be structured to be as minimal as possible. Possible structures could be:

Array:

```json
[
  {
    "ns": "uk.ac.ox.softeng.maurodatamapper.plugins.explorer.research",
    "key": "databaseName",
    "val": "source_ethnicity_code"
  },
  {
    "ns": "uk.ac.ox.softeng.maurodatamapper.plugins.explorer.research",
    "key": "databaseName",
    "val": "source_ethnicity_code"
  }
]
```

Normalised object:

```json
[
  {
    "ns": "uk.ac.ox.softeng.maurodatamapper.plugins.explorer.research",
    "keys": {
      "databaseName": "source_ethnicity_code",
      "targetName": "source_ethnicity_code"
    }
  }
]
```

Flattened keys:

```json
{
  "uk.ac.ox.softeng.maurodatamapper.plugins.explorer.research|databaseName": "source_ethnicity_code",
  "uk.ac.ox.softeng.maurodatamapper.plugins.explorer.research|targetName": "source_ethnicity_code"
}
```

## Convert to Base64

Stringify the JSON object and convert to Base64. This is one reason why the JSON should be as minimal as possible - to produce as small a Base64 string as possible. Base64 is also used because it does not require special URL encoding.

## Include in query string

Include the data in the query string to the Search Listing page e.g.

```
http://localhost:4200/#/search/listing?search=code&labelOnly=true&metadata=ewogICJ1ay5hYy5veC5zb2Z0ZW5nLm1...
```

## Deserialisation

In the Search Listing page, the `metadata` query parameter must then be deserialised to revert the data back to the original format. This can be be used to display the filters on the page (#778) and initiate the search (#781)

pjmonks:
[...]

PR #784, opened 2023-03-09
gh-777 Add profile search filter selection to Search page

### LINKED ISSUE #777: Update Search page to manage profile filters
# Description

The Search page needs to support advanced filter controls to select one or more profile filters to be used as search criteria when searching the catalogue.

A profile filter is a combination of:

1. A _profile name_, which will provide the full _profile namespace_
2. A _metadata key_
3. A _value_ to filter with

# Design

MDM UI Metadata Search-Search page drawio

# Features

The "Advanced" section of the search page should include a set of controls, similar to above, to add, edit and remove profile filters.

Adding a profile filter may need to have a limit imposed, depending on if the UI ends up being too cluttered e.g. a maximum of 3, or 5, profile filters.

Each row should provide:

1. A select control to determine the _profile_ to use. This can be determined via the `GET /api/profiles/providers` endpoint
2. A select control to determine the _metadata key_ to use. The option list must be based on the chosen profile. Use the display name of the key rather than the actual property name.
3. An input control to enter a value.

For entering an input value, two options will currently be supported:

1. If the profile metadata key represents an _enumeration_, the input should be a select control with the possible enumeration options.
2. Otherwise, use a text field input.

# Out of scope

This task should only focus on the UI. Submitting the profile filters to the Search Listing page will be handled in task #779.

pjmonks:
Resolves #777 

- New child component for CatalogueSearchComponent
- Track state of all profile filters using Angular FormArray
- Use profile providers and definitions to guide user for which namespaces and keys to use for filters
- Clean up code in CatalogueSearchAdvancedForm
- Move "More filters" button to parent component
- Section on CatalogueSearch page now controls visibility of the advanced form and profile filter list
- Update code and tests to have @ViewChild mapping work
- Update parent/child communication to handle resets and track validity
- Add and update tests

# Notes

1. Values entered per filter are chosen either from a list of enumeration values or entered as strings. Only "enumeration" field types are especially supported at the moment.
2. The Search page does not pass these chosen filters on to the Search Listing page yet, see #779 

# Screenshots

image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -226,7 +226,7 @@ pipeline {
   post {
     always {
       outputTestResults()
-      zulipNotification(topic: 'mdm-ui')
+      //zulipNotification(topic: 'mdm-ui')
     }
   }
 }
--- a/src/app/catalogue-search/catalogue-search-advanced/catalogue-search-advanced-form.component.html
+++ b/src/app/catalogue-search/catalogue-search-advanced/catalogue-search-advanced-form.component.html
@@ -15,180 +15,171 @@ limitations under the License.
 
 SPDX-License-Identifier: Apache-2.0
 -->
-<div class="container mb-2">
-  <button mat-stroked-button color="primary" (click)="toggleAdvancedSearch()">
-    {{ advancedSearch ? 'Less filters' : 'More filters...' }}
-  </button>
-</div>
-<div *ngIf="advancedSearch">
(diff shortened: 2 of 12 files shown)


abwilson23 (approved):
Code changes are clear and concise. I liked the refactoring of the responsibilities of the parent search component and the individual search components within it. 

Functionality is as expected. Reset and searching both appear to work fine.

PR #788, opened 2023-03-16
gh-459 - Rows disappear from Enumeration table

### LINKED ISSUE #459: Rows disappear from Enumeration table when dragged off of the table
# Description

When adding new enumerations, if you use the drag'n'drop functionality and drop the row away from the table, the row
disappears. 

# Steps to reproduce

1. Create new data element
2. Choose to create a new data type
3. Select "Enumeration"
4. Add two enumeration values
5. Drag one of them off the table

# Expected behavior

Not entirely clear. I believe the default behaviour should be to snap back to the rows original position. 

## Desktop 

Please complete the following information:

- OS: Ubuntu 20 via WSL on Windows 10
- Browser: Chrome
- Version: 97

abwilson23:
Add additional control flow to handle the case where an item is dropped outside of the table. 
In that scenario, just return the table to it's previous state. 

Additionally, remove unused method and renamed drop event.

Fixes #459

### DIFF
--- a/src/app/utility/mc-enumeration-list-with-category/mc-enumeration-list-with-category.component.html
+++ b/src/app/utility/mc-enumeration-list-with-category/mc-enumeration-list-with-category.component.html
@@ -76,7 +76,7 @@ SPDX-License-Identifier: Apache-2.0
     [dataSource]="displayItems"
     cdkDropList
     [cdkDropListData]="dataSource"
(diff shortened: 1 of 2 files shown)


pjmonks (approved):
Changes look simple and effective. You've noted another issue about dragging a row down the table and losing it, that can be a separate issue to fix.

Approved ✔️

Issue #100, opened 2021-05-12
filters don't work on attachments tab


gammonpeter:
filtering doesn't allow me to enter text to filter

PR #794, opened 2023-03-16
gh-761: Theming: Update Api Properties to allow images to be uploaded…

### LINKED ISSUE #761: Theming: Update Api Properties to allow images to be uploaded
# Description

In the future we may add a separate page for theming, but for now we are simply using the Api Properties page to theme the Mdm Explorer. When the explorer.theme.image.header.logo. api property is edited the user needs a page that allows them to upload an image that will be used on the web page.

# Additional context

MauroDataMapper-Plugins/mdm-plugin-explorer#18 will add a new Api Property called explorer.theme.image.header.logo. The value of the explorer.theme.image.header.logo property is to be an image file id.

When the user view the explorer.theme.image.header.logo property in the list of api properties, the value should simply say "Content". (You will see "Content" assigned to other values in the list.) 

When the user goes to the edit page, they should be shown an enhanced page that allows them to view the current image and upload a new image file if desired. Have a look at http://localhost:4200/#/user/profile which allows a user to upload a picture of themselves. This should work in a similar way. The uploaded image will need to be saved to the back end using "theme image file" api calls, (MauroDataMapper/mdm-core#394 will add the required api end points)

NigelPalmer:
… mdm-ui

gh-761

### DIFF
--- a/src/app/admin/theme-image/theme-image.component.scss
+++ b/src/app/admin/theme-image/theme-image.component.scss
@@ -16,21 +16,6 @@ limitations under the License.
 SPDX-License-Identifier: Apache-2.0
 */
 
-.mdm--profile-block {
-  .mdm-profile__image {
-    background-color: rgba(77, 114, 156, 0.1);
-    display: inline-block;
-
-    &--avatar {
-      margin: 16px;
-      width: 256px;
-      height: auto;
-      border-radius: 100%;
-      margin-right: 10px;
-    }
-  }
-}
-
 .mdm-profile__image__edit {
   text-align: right;
 }
--- a/src/app/admin/theme-image/theme-image.component.scss
+++ b/src/app/admin/theme-image/theme-image.component.scss
@@ -1,0 +1,45 @@
+/*
+Copyright 2020-2023 University of Oxford and NHS England
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+.mdm--profile-block {
+  .mdm-profile__image {
+    background-color: rgba(77, 114, 156, 0.1);
+    display: inline-block;
+
+    &--avatar {
+      margin: 16px;
+      width: 256px;
+      height: auto;
+      border-radius: 100%;
+      margin-right: 10px;
+    }
+  }
+}
+
+.mdm-profile__image__edit {
+  text-align: right;
+}
+
+.mdm--background-color {
+    --color-1: #808080;
+    --color-2: #a9a9a9;
+    --size: 8px;
+    
+    background-image: conic-gradient(var(--color-1) 25%, var(--color-2) 25%, var(--color-2) 50%, var(--color-1) 50%, var(--color-1) 75%, var(--color-2) 75%);
+    background-size: calc(var(--size)*2) calc(var(--size)*2);
+}
--- a/src/app/admin/theme-image/theme-image.component.ts
+++ b/src/app/admin/theme-image/theme-image.component.ts
@@ -98,7 +98,7 @@ export class ThemeImageComponent implements OnInit {
   // Saves the theme image
   public saveImage() {
 
-    this.resourcesService.themeImage.update(this.apiPropertyId, { image: this.imageSource, type: 'image/png' }).subscribe((result: { body }) => {
+    this.resourcesService.themeImage.update(this.apiPropertyId, { image: this.imageSource, type: this.getFileType() }).subscribe((result: { body }) => {
       this.messageHandler.showSuccess('Theme image updated successfully.');
       this.imageVersion++;
       this.imageSavedEvent.emit();
@@ -169,5 +169,20 @@ export class ThemeImageComponent implements OnInit {
     else {
       this.value.setValue(this.DefaultImageMessage);
     }
+  }
+
+  private getFileType(): string {
+    if (this.imageSource) {
+      const fileType = this.imageSource.split(':').pop().split(';')[0];
+      if (fileType) {
+        if (fileType === 'image/jpg') {
+          return 'image/jpeg'
+        }
+        else {
+          return fileType
+        }
+      }  
+    }
+    return 'unknown'
   }
 }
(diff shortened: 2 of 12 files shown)


NigelPalmer:
MauroDataMapper/mdm-resources#100 is required to be merged into develop so this will build.

src/app/admin/theme-image/theme-image.component.scss
```diff
@@ -0,0 +1,45 @@
...
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+.mdm--profile-block {
+  .mdm-profile__image {
+    background-color: rgba(77, 114, 156, 0.1);
+    display: inline-block;
+
+    &--avatar {
```
pjmonks: I'm assuming the `mdm-theme-image` component was copied from something else, this CSS class isn't used and is superfluous. Other styles may also not be required either, please double check.

src/app/admin/theme-image/theme-image.component.ts
```diff
@@ -0,0 +1,188 @@
...
+  selector: 'mdm-theme-image',
+  templateUrl: './theme-image.component.html',
+  styleUrls: ['./theme-image.component.scss']
+})
+export class ThemeImageComponent implements OnInit {
+  @Input() apiPropertyId: Uuid;
+  @Output() imageSavedEvent = new EventEmitter();
+  @Output() imageChangedEvent = new EventEmitter<ImageChangedEvent>();
+
+  currentUser: any;
```
pjmonks: According to the constructor when `securityHandler.getCurrentUser()` is used, this should be `currentUser: UserDetails`

src/app/admin/theme-image/theme-image.component.ts
```diff
@@ -0,0 +1,188 @@
...
+  imageVersion = 1;
+  imageSource: any = '';
+  themeImagePath: string;
+  trustedUrl: any;
```
pjmonks: Rather than `any`, this seems to be a `SafeResourceUrl` type.

src/app/admin/theme-image/theme-image.component.ts
```diff
@@ -0,0 +1,188 @@
...
+  styleUrls: ['./theme-image.component.scss']
+})
+export class ThemeImageComponent implements OnInit {
+  @Input() apiPropertyId: Uuid;
+  @Output() imageSavedEvent = new EventEmitter();
+  @Output() imageChangedEvent = new EventEmitter<ImageChangedEvent>();
+
+  currentUser: any;
+  imageVersion = 1;
+  imageSource: any = '';
```
pjmonks: Think this should be `imageSource: string`

src/app/admin/theme-image/theme-image.component.ts
```diff
@@ -0,0 +1,188 @@
...
+  themeImagePath: string;
+  trustedUrl: any;
+  afterSave: (result: { body: { id: any } }) => void;
```
pjmonks: This isn't referenced anywhere so can be removed.

src/app/admin/theme-image/theme-image.component.ts
```diff
@@ -0,0 +1,188 @@
...
+        })
+      )
+      .subscribe((data: ApiPropertyResponse) => {
+        this.showImage = this.isUUID(data.body.value);
+        this.showDefaultMessageIfRequired();
+        this.originalValueWasDefault = (this.value?.value === this.DefaultImageMessage);
+      });
+  }
+
+  // Saves the theme image
```
pjmonks: Comment is superfluous, the function name is clear enough I think.

src/app/admin/theme-image/theme-image.component.ts
```diff
@@ -0,0 +1,188 @@
...
+  fileChangeEvent(fileInput: any): void {
+    this.readThis(fileInput.target);
+    this.showImage = true;
+    this.showDefaultMessageIfRequired();
+    this.imageChangedEvent.emit({
+      changeEvent: ImageChangeType.uploaded
+    });
+  }
+
+  // When an image is removed
```
pjmonks: More unecessary comments I feel. Maybe also check the other functions.

src/app/admin/theme-image/theme-image.component.ts
```diff
@@ -0,0 +1,188 @@
...
+    return (
+      value.match(
+        '^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$'
+      ) !== null
+    );
+  }
+
+  private showDefaultMessageIfRequired() {
+    if (this.showImage) {
+      // This is required to avoid a mat-form-field must contain a MatFormFieldControl error
```
pjmonks, line 184: This is an example of a good comment in comparison, since it explains something not obvious 👍

src/app/admin/theme-image/theme-image.component.ts
```diff
@@ -0,0 +1,188 @@
...
+      this.value.setValue(' ');
+    }
+    else {
+      this.value.setValue(this.DefaultImageMessage);
+    }
+  }
+
+  private getFileType(): string {
+    if (this.imageSource) {
+      const fileType = this.imageSource.split(':').pop().split(';')[0];
```
pjmonks, line 195: Nothing wrong with the code here, I'd only suggest a comment to explain why it is determining the file type this way - what is the expected format of the `imageSource` string? Why split on these characters? etc

pjmonks (changes requested):
The UI changes look good, it looks nice how the image you are uploading is displayed and automatically handled within the context of it being an API property also.

Only feedback I've given is some minor comments on tidying up the new `theme-image.component`.

src/app/admin/theme-image/theme-image.component.scss
abwilson23: Removed.

src/app/admin/theme-image/theme-image.component.ts
abwilson23: Added correct type.

src/app/admin/theme-image/theme-image.component.ts
abwilson23: Additionally, the `public` qualifier is unnecessary because methods in typescript classes default to being public.

src/app/admin/theme-image/theme-image.component.ts
abwilson23: Changed. And added a type check prior to reading the result of the FileReader.

src/app/admin/theme-image/theme-image.component.ts
abwilson23: Changed.

src/app/admin/theme-image/theme-image.component.ts
abwilson23: Removed.

src/app/admin/theme-image/theme-image.component.ts
abwilson23: Checked and removed unnecessary comments.

src/app/admin/theme-image/theme-image.component.ts
abwilson23: Added a clarifying comment.

pjmonks (dismissed):
@abwilson23 has implemented the feedback on behalf of @NigelPalmer

pjmonks (approved):

PR #795, opened 2023-03-16
gh-331: avoid graphs to start too close to controls

### LINKED ISSUE #331: The default layout of the Versioning graph page has bigger resolution.
**Describe the bug**
When a user log into Mauro to view and track activities of a model, the versioning graph page opening by default has a bigger resolution, where the information on the graph are very close to the reset functions on the page. 
**To Reproduce**
Steps to reproduce the behaviour:
1.	Go to 'Mauro'
2.	Click on 'a Model 
3.	Select the <three dots> close to the <user & group access> menu
4.	Keep Focus on the <Merge> menu and select the <Show merge graph>
5.	“Versioning graph” page is displayed 
6.	Verify the resolution layout of the content
**Expected behaviour**
The page should display the contents well-presented and not too close to the function menus
**Screenshots**
If applicable, add screenshots to help explain your probl
Merge layout
em.

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

CristianGimenez:
After playing around and reading a lot of documentation about jointjs I have not found a clean good way of setting an offset or padding to the paper. Tried to set origin to different coordinates as their [online example ](https://www.jointjs.com/demos/paper-attributes) seems to suggest that could do the trick. Nothing of what I tried has worked so I ended up with 2 alternatives:
1. Adding a rectangle on the top left to avoid anything else to occupied that space.
2. Start the diagram with lower zoom.

Of those 2 options, I think option 2 is the best. It seems to solve the problem, but please, I'd love to hear any other suggestions.

Fixes #331

### DIFF
--- a/src/app/diagram/diagram/diagram.component.ts
+++ b/src/app/diagram/diagram/diagram.component.ts
@@ -58,7 +58,6 @@ export class DiagramComponent implements OnInit {
   isLoading: boolean;
   isPopup = false;
 
-
   initPan: SvgPanZoom.Point;
   initZoom: number;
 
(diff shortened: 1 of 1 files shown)


pjmonks (approved):
I updated to fix some eslint issues (which also updated a lot of code formatting), but the fix you made I tested and am happy that it works 👍

PR #796, opened 2023-03-17
gh-792 Fix ordering of enumeration values

### LINKED ISSUE #792: Items disappear from enumeration options table when moved from a lower to a higher position in the list.
# Description

When dragging an element from a lower index to a higher index (vertically higher to lower) in the enumerations options list 
the element disappears! From lower to higher correctly rearranges the list as expected. 

# Steps to reproduce

1. Go to a data class
2. Create a new data element, choose type enumeration.
3. Add more than one option to the list.
4. Attempt to drag an item from a lower to a higher position
5. Witness paranormal activity

# Expected behavior

The item to take it's new position in the list. 

## Desktop 

Please complete the following information:
- OS: WSL2 on Windows 11
- Browser: Chrome
- Version: 111

pjmonks:
Client side ordering (for creating new enumerations):

- Track items being swapped correctly
- Don't do manual array handling, use same CDK function as handleDrop()
- No longer loses last row if dragging down the table

Server side ordering (for existing enumerations):

- Use correct index, was using 1-based starting index originally
- Clean up server side code updates

Fixes #792

### DIFF
--- a/src/app/utility/mc-enumeration-list-with-category/mc-enumeration-list-with-category.component.ts
+++ b/src/app/utility/mc-enumeration-list-with-category/mc-enumeration-list-with-category.component.ts
@@ -36,10 +36,20 @@ import { StateHandlerService } from '@mdm/services';
 import { EditingService } from '@mdm/services/editing.service';
 import {
   DataClassDetail,
-  DataModelDetail
+  DataModelDetail,
+  DataTypeDetailResponse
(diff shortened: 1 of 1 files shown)


abwilson23 (approved):
Code changes are concise and easy to understand. This fixes the ordering issue and 
the side issue of server side ordering. ✅ 

Just a note that the behaviour when moving items between different groups is still strange, however that is outsie the scope of this ticket.

PR #797, opened 2023-03-20
gh-624 Track changed state in bulk editor tabs

### LINKED ISSUE #624: Show leave confirmation message at appropriate times in bulk editor
# Description

The current implementation of the bulk editor will always show a confirmation message if you choose to leave and navigate to a different page:

image

This confirmation logic could be improved by only showing if there are unsaved changes - it seems pointless showing the message if you have definitely saved all your current work. This would most likely require better tracking of edit changes made in each tab/spreadsheet.

pjmonks:
- Confirm switching panes or closing if values have changed but not been saved
- Also fixed bug with auto sizing columns due to wrong column key used

Resolves #624 

To test:

1. Bulk edit a model
2. Make your item/profile selections to view the editor tabs
3. Try closing or going back a step - both when you've made changes in each tab and when you haven't. A confirmation dialog should appear in the correct scenarios.

### DIFF
--- a/src/app/bulk-edit/bulk-edit-container/bulk-edit-container.component.ts
+++ b/src/app/bulk-edit/bulk-edit-container/bulk-edit-container.component.ts
@@ -22,7 +22,7 @@ import { MauroItem } from '@mdm/mauro/mauro-item.types';
 import { StateHandlerService, MessageHandlerService } from '@mdm/services';
 import { EditingService } from '@mdm/services/editing.service';
 import { UIRouterGlobals } from '@uirouter/core';
-import { EMPTY } from 'rxjs';
+import { EMPTY, of } from 'rxjs';
 import { catchError, filter } from 'rxjs/operators';
 import { BulkEditContext, BulkEditStep } from '../bulk-edit.types';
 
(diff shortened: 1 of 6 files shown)


src/app/bulk-edit/bulk-edit-container/bulk-edit-container.component.ts
```diff
@@ -78,20 +79,38 @@ export class BulkEditContainerComponent implements OnInit {
...
+    });
+  }
+
+  onChanged() {
+    this.hasChanged = true;
+  }
+
+  onSaved() {
+    this.hasChanged = false;
   }
```
abwilson23, line 115: Not necessary to change, but wondering if observables could/should be used instead of these setters. Something like, create a `hasChangedEvent$` stream which listens to a combined stream of the changed/saved button click events in the child component and uses the latest emission to set the value of `hasChanged`. mainly I'm just curious if you have any thoughts about that as an alternative method to traditional event bindings. Certainly, this way appears simpler/easier to grok at a glance.

abwilson23 (approved):
Changes are clear and functionality is as expected.

Issue #102, opened 2021-05-12
Terminology XML export fails due to native element error


gammonpeter:
Click on a Terminology then click on export and select xml, using dev tool look at errors

PR #798, opened 2023-03-20
gh-781 - add profileFilters to searchQuery

### LINKED ISSUE #781: Send profile filters in Mauro search request
# Description

Once #779 and MauroDataMapper/mdm-core#399 are completed, all profile filters should be passed to the Mauro search endpoint for the backend to process.

# Parameters

The `CatalogueSearchService` will need to be updated to send this information in the `POST` request:

```json
{
  // ... other search parameters ...
  "profileFields": [
    {
      "metadataNamespace": "...",
      "metadataPropertyName": "...",
      "filterTerm": "...",
      "type": "phrase"  // phrase | keyword (default: phrase)
    }
  ]
}
```

Send one or more filters inside the `profileFields` array. Each one will have:

1. The namespace from the profile
2. The metadata key
3. The value
4. Whether to use `phrase` or `keyword` search.

The type should be determined as:

1. If the metadata key represents an enumeration, use `phrase` for exact matching
2. Otherwise, use `keyword` for fuzzy matching

abwilson23:
Retrieve profileFilter information prior to firing off catalogue search request. A relatively straightforward transform of the values using an existing utility function that's already well tested hence why I haven't added additional tests. 

This is partnered with this PR: https://github.com/MauroDataMapper/mdm-resources/pull/102
which needs to be merged _before_ the PR checks will pass on this PR.

### DIFF
--- a/src/app/catalogue-search/catalogue-search.service.ts
+++ b/src/app/catalogue-search/catalogue-search.service.ts
@@ -24,13 +24,14 @@ import {
   PageParameters,
   ProfileDefinition,
   ProfileDefinitionResponse,
+  ProfileFieldQueryData,
   ProfileSummary,
   ProfileSummaryResponse,
   SearchQueryParameters
 } from '@maurodatamapper/mdm-resources';
 import { MdmResourcesService } from '@mdm/modules/resources';
 import { forkJoin, Observable, of } from 'rxjs';
-import { map } from 'rxjs/operators';
+import { map, switchMap } from 'rxjs/operators';
 import {
   CatalogueSearchContext,
   CatalogueSearchParameters,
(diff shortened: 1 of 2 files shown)


src/app/catalogue-search/catalogue-search.service.ts
```diff
@@ -53,20 +54,55 @@ export class CatalogueSearchService {
...
+              page,
+              items: searchResults.items
+            };
+          })
+        );
+      })
+    );
+  }
+
+  private getProfileFiltersForQuery(
```
CristianGimenez, line 87: I was expecting eslint to complain about a private function in between non private functions, but it did not. So, I guess this is all right.

CristianGimenez (approved):
Everything looks good to me (great renaming "since you were there"). I've ran pr-checks and I haven't seen any failing tests nor new eslint warnings/errors. @abwilson23 could you update the branch with the latest develop changes? That will hopefully trigger the tests to run again, and now the mdm-resources ticket is merged they should pass and we should be ready to merge.

PR #800, opened 2023-03-23
gh 718 - Replace use of ag-grid packages with modules.  

abwilson23:
This PR was borne out of an attempt to decrease bundle size via lazy loading. That investigation led to some possible alternatives, one of them being switching to using ag-grid _modules_ rather than _packages_. The impact ended up being smaller than hoped, but it's a start. The original ticket has been updated with some additional information here: https://github.com/MauroDataMapper/mdm-ui/issues/718

Here you can see the bundled **package** and it's size.
image

and here you can see the bundled **modules** and their size. 
image

Wow! 3 kB improvement.. 😮 

The code changes are minor and just reflect switching some imports and adding a type to meet a type requirement. One small note is that `AgModule.withComponents` is no longer needed due to updates in the library. 

**Note** I removed a console.log in the activeSessions file and prettier has gone and reformatted everything. The only change is the console.log.


package.json
```diff
@@ -24,6 +24,10 @@
   },
   "private": true,
   "dependencies": {
+    "@ag-grid-community/angular": "^29.2.0",
```
pjmonks: These package references should not include the "^" symbol. By not not including, it will always lock `npm install` to those exact versions, allowing us to better control which versions get installed. It is also something that RenovateBot would attempt to fix anyway.

src/app/app.component.ts
```diff
@@ -36,6 +36,10 @@ import { FeaturesService } from './services/features.service';
 import { SharedService } from './services/shared.service';
 import { ThemingService } from './services/theming.service';
 import { FooterLink } from './shared/footer/footer.component';
+import { ModuleRegistry } from '@ag-grid-community/core';
+import { ClientSideRowModelModule } from '@ag-grid-community/client-side-row-model';
+
+ModuleRegistry.registerModules([ClientSideRowModelModule]);
```
pjmonks: This seems relevant to getting agGrid to initialise, but is there a better place to put this? For example, since we so far only use agGrid for bulk editing, could this be set in the `BulkEditModule`? It would then be closer to where it is used, at the moment someone might have trouble tracking down why this line of code is here in the app root itself.

pjmonks (changes requested):
My testing shows the bulk editor works the same. Although it doesn't achieve much of an impact on the original goal of reducing app bundle size, keeping the agGrid components more up to date it also not unreasonable.

Could you check the minor comments I left please before I merge?

package.json
abwilson23: Fixed.

src/app/app.component.ts
abwilson23: Definitely, this was a relic of testing and me forgetting to commit the change to the bulk-edit module... fixed!

abwilson23:
Requested changes have been made 👍

pjmonks (approved):
All looks good 👍

PR #801, opened 2023-03-27
gh-573 Update built-in Markdown editor

### LINKED ISSUE #573: Include toolbar options in the markdown editor
# Description

The HTML editor uses a 3rd party library called [Jodit](https://xdsoft.net/jodit/) to provide a WYSIWYG editor experience for entering HTML. The Markdown editor however does not provide much in the way of supporting editing options. There is a link to a help page for markdown syntax, but it leaves the entry of text up to the user.

One way to improve this is to include custom toolbar options to handle the basic markdown syntax - such as inserting bold/italic text, headings etc.

# Additional context

Markdown syntax is not very complex, so I believe the only necessary options to include would be:

* Bold - to enter `**text**`
* Italic - to enter `_text_`
* Headings - to enter various heading levels - `#`, `##`, etc
* Quote - to enter `> text`
* Bullet and numbered lists

There are currently three existing toolbar buttons in our markdown editor already, these can remain and be supplemented with the actions above.
 
image

Another alternative could be to search for a 3rd party WYSIWYG markdown editor, but this may be overkill since the possible actions above are not very complicated.

pjmonks:
Resolves #573 

**Note:** despite requesting multiple reviewers, I think @jamesrwelch should have the final say on how the new editor UI looks/functions. Other reviewers may leave feedback on the code changes.

- Move HTML and Markdown controls to a new content module
- Refactor markdown-text-area component
  - Clean up unused code
  - Simplify insert text functionality
- Add new toolbar with common Markdown actions. Supports:
  - Bold
  - Italic
  - Heading
  - Block quote
  - Bullet and numbered lists
- Improve link dialog selection and text insertion
- Fix style of blockquotes
- Change markdown preview to swap editor and preview panes
- Add keyboard shortcuts
- Add placeholder text
- Add new tests to verify some of the editing functions of the Markdown editor

# Screenshots

image

image


jamesrwelch:
Sorry, @pjmonks - would you be able to resolve the conflicts on this one again please?  I promise I'll review it before any more PRs are accepted!

pjmonks:
@jamesrwelch Merge conflicts have now been resolved.

jamesrwelch:
This is a really good piece of work.  Some minor comments as I'm going through:
- The preview pane has a max-height but is usually a different size to the edit pane.  This makes it quite choppy when switching between preview and edit.  Is it possible to make the boxes the same size?
- In the editor on a Mac, I get Bold by Ctrl+B, but in most other Mac applications it's Cmd+B.  Is it possible to change that?
- If you try to bold something multiple times, it doesn't toggle but adds multiple stars, making the markdown invalid.  I'm not sure what the right thing to do here is
- There seems to be an inconsistency in using the toolbar actions.  If I select some text and click 'bold', it puts `**` around it - which is correct.  If I select some lines and click 'numbered list', it leaves the selected text unchanged and inserts a new placeholder.  
- We've got a 'link to catalogue item' option, which is great - could we also have a 'add hyperlink' for a normal external link?
-

pjmonks:
@jamesrwelch Thanks for this feedback.

For two of your points - toggle/untoggle bold and inline vs block insertion - I should point out this is very basic/minimal text insertion. I looked at some source code for Markdown editors you could install instead and they all have some level of "intelligence" as to what is currently entered, specifically state tracking - is the cursor within a bold statement or not, what current numbered item in a list you're in etc. This is too complex for me to look into at the moment so I chose to do plain text insertion and hope the user could at least see the syntax each command uses to identify what they should do. Not ideal, but otherwise I see either a new task to fix these or look into a 3rd party library.

Your other 3 points I think I can easily put in though - preview pane max-height, correct keyboard shortcuts and insert a standard link. For the keyboard shortcuts, it would be useful to me if you gave me a list of what you would expect to use on a Mac for each command since I don't use a Mac myself.

jamesrwelch:
> For the keyboard shortcuts, it would be useful to me if you gave me a list of what you would expect to use on a Mac for each command since I don't use a Mac myself.

The same as you have, but using `Cmd` rather than `Ctrl`.  I think this might help:  https://stackoverflow.com/questions/3902635/how-does-one-capture-a-macs-command-key-via-javascript

pjmonks:
Updates after feedback:

1. Preview pane has a minimum height now
2. New command added - insert standard link `[link text](http://url)`
3. Updated the keyboard shortcuts to also make sense on Mac OS too - @jamesrwelch please confirm these work as I cannot test them

Keyboard shortcuts are:

```
Mac OS:

'Bold (⌘ + B)',
'Italic (⌘ + I)',
'Heading (⌘ + H)',
'Quote (⌘ + \')',
'Numbered list (⌘ + Shift + L)',
'Bullet list (⌘ + L)',
'Link to URL (⌘ + K)',
'Link to catalogue element (⌘ + Shift + K)'

Other:

'Bold (Ctrl + B)',
'Italic (Ctrl + I)',
'Heading (Ctrl + H)',
'Quote (Ctrl + \')',
'Numbered list (Ctrl + Shift + L)',
'Bullet list (Ctrl + L)',
'Link to URL (Ctrl + K)',
'Link to catalogue element (Ctrl + Shift + K)'
```

jamesrwelch (approved):
Thanks @pjmonks - this seems to work nicely.  
The keyboard shortcuts now work on a Mac, thank you.  
The code looks clean.
At some point we'll need to work out how to rationalise the interaction between HTML and Markdown, but that's not a problem for this PR.
Approved!

Issue #104, opened 2021-05-13
Deleting SourceFlowDataModel Throws exception 


gammonpeter:
Using the Bootstrap dataclasses delete SourceFlowDataModel the following error is returned

```{
  "headers": {
    "normalizedNames": {},
    "lazyUpdate": null
  },
  "status": 500,
  "statusText": "OK",
  "url": "http://localhost:8080/api/dataModels/6aaf8e1e-b91e-415c-9d86-e4fc2ccc87f8?permanent=true",
  "ok": false,
  "name": "HttpErrorResponse",
  "message": "Http failure response for http://localhost:8080/api/dataModels/6aaf8e1e-b91e-415c-9d86-e4fc2ccc87f8?permanent=true: 500 OK",
  "error": {
    "status": 500,
    "reason": "Internal Server Error",
    "errorCode": "UEX--",
    "message": "No signature of method: uk.ac.ox.softeng.maurodatamapper.dataflow.component.DataElementComponentService.deleteAllFacetsByCatalogueItemIds() is applicable for argument types: (ArrayList, String) values: [[24a4dd67-2548-4acd-bd2c-0ba977649a69, 32772381-b2c8-44ee-b1a4-1d2fee0087aa, ...], ...]",
    "path": "/api/dataModels/6aaf8e1e-b91e-415c-9d86-e4fc2ccc87f8",
    "devMode": true,
    "version": "4.5.0-SNAPSHOT",
    "exception": {
      "type": "MissingMethodException",
      "message": "No signature of method: uk.ac.ox.softeng.maurodatamapper.dataflow.component.DataElementComponentService.deleteAllFacetsByCatalogueItemIds() is applicable for argument types: (ArrayList, String) values: [[24a4dd67-2548-4acd-bd2c-0ba977649a69, 32772381-b2c8-44ee-b1a4-1d2fee0087aa, ...], ...]",
      "stacktrace": [
        "uk.ac.ox.softeng.maurodatamapper.dataflow.component.DataElementComponentService.$tt__dataElementComponentService_deleteAllByModelId(DataElementComponentService.groovy:123)",
        "jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)",
        "jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)",
[...]

PR #802, opened 2023-04-03
Fix issue when creating dataTypes

### LINKED ISSUE #318: Adding new Data Type updates existing Data Type
**Describe the bug**
When editing an Element in a Class, clicking on 'Add a new Data Type' updates the Element's current Data Type to the selected values, rather than creating a new Data Type.

**To Reproduce**

Below steps are for example in development version.

1. Go to Complex Test Data model and add two integer elements as below:
Screen Shot 2021-09-16 at 15 42 45

2. Click Add a new Data Type and select a Reference to the content class. N.B. some red error text appears which is probably not necessary:
Screen Shot 2021-09-16 at 15 49 49
Screen Shot 2021-09-16 at 15 45 49

3. Then the integer data type is updated to be a content class reference:
Screen Shot 2021-09-16 at 15 45 26
Screen Shot 2021-09-16 at 16 13 43

**Expected behavior**
New Data Type should be added and only the selected Element being edited should have its Type changed.

**Versions**
Reproduced on Mauro version 6.6.0 (in mdm-docker) and version 6.7.0-SNAPSHOT (development version).

CristianGimenez:
This PR will be marked as "Draft" because it depends on a [PR for mdm-resources](https://github.com/MauroDataMapper/mdm-resources/pull/104).

The main cause of the issue was that `new-data-type-inline.component` was using an anonymous type with the wrong variable names, and `default-profile-editor-modal.component` was using those variable names without treating them. All this is allowed because the `DataType`definition allows `[key : string]: value` as a variable.

There is a number of changes in this PR:
- Removing unused validations and some small fixes that I spotted while working in the different components.
- Removing anonymous types to use `DataType`.
- Adding some comments on the code to describe better what needs to happen or why we do certain transformations.
- When creating a new `DataType` the label for the new `dataType `is mandatory, so we should show the fields from the start.

Both creating a new `dataType` while editing an existing `dataElement `and when creating a brand new `dataElement `should work fine now.

closes #318

### DIFF
--- a/src/app/data-type/data-type.component.ts
+++ b/src/app/data-type/data-type.component.ts
@@ -26,10 +26,16 @@ import { EditingService } from '@mdm/services/editing.service';
 import {
   ElementTypesService,
   MessageHandlerService,
-  SecurityHandlerService,
+  SecurityHandlerService
 } from '@mdm/services';
 import { DefaultProfileItem } from '@mdm/model/defaultProfileModel';
-import { DataModel, DataType, DataTypeDetail, DataTypeDetailResponse, Uuid } from '@maurodatamapper/mdm-resources';
+import {
+  DataModel,
+  DataType,
+  DataTypeDetail,
+  DataTypeDetailResponse,
+  Uuid
+} from '@maurodatamapper/mdm-resources';
 import { TabCollection } from '@mdm/model/ui.model';
 import { BaseComponent } from '@mdm/shared/base/base.component';
 
@@ -38,7 +44,8 @@ import { BaseComponent } from '@mdm/shared/base/base.component';
   templateUrl: './data-type.component.html',
   styleUrls: ['./data-type.component.scss']
 })
-export class DataTypeComponent extends BaseComponent
+export class DataTypeComponent
+  extends BaseComponent
   implements OnInit, AfterViewInit {
   @ViewChild('tab', { static: false }) tabGroup: MatTabGroup;
 
@@ -84,7 +91,7 @@ export class DataTypeComponent extends BaseComponent
     private messageHandler: MessageHandlerService,
     private securityHandler: SecurityHandlerService,
     private elementTypes: ElementTypesService,
-    private editingService: EditingService,
+    private editingService: EditingService
   ) {
     super();
   }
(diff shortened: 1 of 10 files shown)


src/app/dataElement/data-element/data-element.component.ts
```diff
@@ -221,7 +215,20 @@ export class DataElementComponent
         resource.minMultiplicity = item.minMultiplicity as number;
         resource.maxMultiplicity = item.maxMultiplicity;
       } else if (item.controlType === ProfileControlTypes.dataType) {
-        resource.dataType = item.value as DataTypeReference;
+        const castedValue = item.value as DataType;
```
abwilson23: Not a necessary change but given we've identified that `item.controlType` is `datatype` a more descriptive name than `castedValue` might be `dataTypeValue`.

src/app/data-type/data-type.component.ts
```diff
@@ -129,8 +132,6 @@ export class DataTypeComponent extends BaseComponent
 
         this.dataType.classifiers = this.dataType.classifiers || [];
         this.loadingData = false;
-        this.showExtraTabs =
-          !this.sharedService.isLoggedIn() || !this.dataType.editable;
```
abwilson23, line 133: Just a note that you can remove the injected SharedService from the the constructor as this was the only use of it.

abwilson23 (dismissed):
Couple of small comments, but overall code changes appear sensible and concise. The functional testing went well, changing data types and adding new ones has the described and desired behaviour. 

It appears there are some jenkins tests to sort out still before it can be merged.

src/app/dataElement/data-element/data-element.component.ts
CristianGimenez: Makes sense, changing.

src/app/data-type/data-type.component.ts
CristianGimenez: Done

CristianGimenez:
@abwilson23 I've addressed your comments in the latest commit. Could you resolve the comments if you are happy? Whatever Jenkins was unhappy about seems to be ok now.

abwilson23 (approved):
Approved ✅

jamesrwelch:
@abwilson23 @CristianGimenez I'm happy to merge this now, if you're both happy?

CristianGimenez:
> @abwilson23 @CristianGimenez I'm happy to merge this now, if you're both happy?

Yes please, thanks!

jamesrwelch:
Fantastic work both of you 👍

PR #804, opened 2023-04-20
gh-427 Correctly route catalogue items to views using Mauro paths

### LINKED ISSUE #427: Update the way links in descriptions work to use the new 'Path' functionality
## Introduction

Any item (container, model, model item, facet) in Mauro can now be referenced by a path.  This ticket is to update the link functionality in descriptions and other long text fields to use the new paths for links between Mauro items.  The aim is that these name-based paths can be used instead of UUIDs in links, so that descriptions are more easily portable and can be more easily created automatically during bulk operations.

- This applies to descriptions, and to any other long text field in profiles, etc where the HTML / Markdown editors are used.
- It should be implemented in both HTML and Markdown editors

It is related to #417 which provides an alternative to the current wizard for choosing an element.

## Paths

A path in Mauro can take one of two forms: _absolute_ and _relative_

- _Absolute_ paths start from a folder or model identified by name and version.
- _Relative_ paths start from the containing model of the item that is being edited.  

A path has the syntax:

`type : name [version] | more_path`

Where: 
- `type` denotes the type of Mauro item - for example `dm` is DataModel, `fo` is Folder, etc.  Types are always two characters
- `name` is the label of the item in question.
- `version` is used to disambiguate finalised models and versioned folders.  If no `version` is provided, then the latest model with that label will be chosen.

The `|` character is used to build up chains of paths to navigate through the hierarchy.

### Examples

`dm:DataModel1` is a link to the latest version of the model named ‘DataModel1’.  Note you don’t need the folder path to reach a data model or versioned folder.
`dm:DataModel1|dc:DataClass1` is a link to the top-level DataClass called ‘DataClass1’ in the latest version of the data model above.

`dc|DataClass1` is a relative link to the top-level DataClass within the current model.

## Functionality

- When a long text field is rendered - either normally, or in ‘preview’ mode when editing (in HTML, and in Markdown), the UI should be able to resolve those links. That may involve adding a prefix to the urls to ensure that they lead somewhere within the single-page app.  This can use the Pathing API endpoints to resolve to a Mauro item.
- When a long text field is edited, the links should be displayed using this pathing notation so that they can be easily edited by users
- When the ‘add element’ button is selected, the selection of a Mauro item should result in a link that uses this pathing notation.

## Decisions

- We will need to decide on a syntax to disambiguate absolute and relative paths.  For example, a path starting `dm:DataModel1` might either be an absolute path to a data model, or a relative path within the same versioned folder.  Initial suggestion would be a prefix of `||` for an absolute path (c.f. xpath), but happy to take suggestions.
- When links are manually added through the item picker, we might need a way to decide whether an absolute or default path is required.  My suggestion is that relative paths should be used where possible, as this ensures links can be resolved when models are imported into other catalogues.

pjmonks:
**Note:** this PR depends on MauroDataMapper/mdm-resources#105 to be merged first.

Resolves #427 

Updates the Markdown and HTML editor catalogue item link system to output a standardised redirect URL in the Angular routes. The redirect URL then uses custom logic to locate the catalogue item using its path and render the correct view to display it.

- Create redirect state in UI Router to go to a specific catalogue item by path
- Create map of path prefixes to pathable domain types
- Mark all necessary routes with domain types to match against catalogue items
- Fix Markdown insertion and rendering of Mauro paths
  - Insert Markdown link using catalogue item path returned from Mauro as the href part
  - Adjust Markdown renderer to map paths to the single route which redirects all catalogue items based on path
  - Fix Markdown rendering of links using more accurate regex to identify hrefs that also contain brackets
- Update HTML editor to use correct Mauro path links
- Remove obsolete LinkCreatorService


joe-crawford (approved):
Thanks for all your work on this and the linked changes @pjmonks.

Have tested locally and adding links to pathed CatalogueItems using the Markdown editor is working really well and the navigation to the links looks good as well. 🎉

PR #805, opened 2023-04-25
gh-778: Filters in search page

### LINKED ISSUE #778: Update Search Listing page to display profile filters
# Description

The Search Listing page needs to be updated to support listing chosen profile filters from the previous Search page.

# Design

MDM UI Metadata Search-Search listings page drawio

# Features

Once a list of profile filters have been identifier, a simple list of them should appear in the "Filters" column of the Search Listing page. Each profile filter should display:

1. The name of the profile
2. The display name of the metadata key
3. The value chosen

It will not be possible to edit these profile filters directly from the page. However, a "remove" button will be next to each one. Removing a profile filter should trigger an update search and render of the page.

The section listing the profile filters will also include an "Add" or "Modify" button for altering these profile filters - see #780.

# Out of scope

This task will not handle:

- Modifying profile filters from the Search Listing page - see #780
- Loading the correct filters to display on the page - see #779
- Running the search with the displayed filters - see #781

### LINKED ISSUE #780: Modify profile filters from the Search Listing page
# Description

Once #778 is completed, the ability to modify the selected profile filters from the Search Listing page should be added.

# Design

MDM UI Metadata Search-Search listings metadata filters drawio

# Features

Editing the profile filters inline on the page would be too cumbersome. Instead, a dialog window should be launched to list the profile filters with more space. The same controls or parent component from #777 could be reused to display in this dialog.

Once changes are made and the dialog is closed, the new selection should be updated back onto the page and the search should be run again.

OButlerOcc:
Added new components for handling the profile filters on the search page.
Integrated them in the search listing page, had to make some minor changes elsewhere to accommodate.

Resolves #778 
Resolves #780


src/app/catalogue-search/profile-filter-dialog-component/profile-filter-dialog-component.component.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { ProfileFilterDialogComponent } from './profile-filter-dialog-component';
+
+describe('ProfileFilterDialogComponent', () => {
+  let component: ProfileFilterDialogComponent;
+  let fixture: ComponentFixture<ProfileFilterDialogComponent>;
+
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({
+      declarations: [ProfileFilterDialogComponent]
```
pjmonks: You have some tests failing here due to missing dependencies. For now you can these to make the tests pass minimally:

```ts
await TestBed.configureTestingModule({
      declarations: [ProfileFilterDialogComponent],
      providers: [
        {
          provide: MatDialogRef,
          useValue: jest.fn()
        },
        {
          provide: MAT_DIALOG_DATA,
          useValue: jest.fn()
        }
      ]
    }).compileComponents();
```

src/app/catalogue-search/catalogue-search-listing/catalogue-search-listing.component.html
```diff
@@ -76,6 +76,16 @@
           (filterChange)="onFilterChanged($event)"
           (filterReset)="onFilterReset()"
         ></mdm-search-filters>
+        <mat-divider></mat-divider>
+        <div>
```
pjmonks: The surrounding `<div>` can be removed, this isn't necessary

src/app/catalogue-search/catalogue-search-listing/catalogue-search-listing.component.ts
```diff
@@ -63,7 +66,8 @@ export class CatalogueSearchListingComponent implements OnInit {
     private routerGlobals: UIRouterGlobals,
     private stateRouter: StateHandlerService,
     private catalogueSearch: CatalogueSearchService,
-    private messageHandler: MessageHandlerService
+    private messageHandler: MessageHandlerService,
+    private profileFilterDialog: MatDialog
```
pjmonks: This member name can be generalised to just `dialog` - the `MatDialog` service can open any dialog component.

src/app/catalogue-search/catalogue-search-listing/catalogue-search-listing.component.ts
```diff
@@ -175,6 +179,40 @@ export class CatalogueSearchListingComponent implements OnInit {
...
+    }
+    this.updateSearch();
+  }
+
+  openProfileFilterDialog() {
+    const dialogRef = this.profileFilterDialog.open(
+      ProfileFilterDialogComponent,
+      {
+        width: '95vw',
+        height: '15vw',
```
pjmonks: These dialog dimensions are currently fixed and don't react very well to changes in screen size. After experimenting, I'd suggest changing to these:

```ts
{
  minWidth: '70vw',
  maxHeight: '40vw'
}
```

The dialog then seems to look more fitted around the content and will grow and be limited by height appropriately.

src/app/catalogue-search/catalogue-search-profile-filter-list/catalogue-search-profile-filter-list.component.ts
```diff
@@ -55,12 +58,15 @@ export class CatalogueSearchProfileFilterListComponent
 
   private unsubscribe$ = new Subject<void>();
 
+  constructor(
+    private resources: MdmResourcesService,
+    public cdr: ChangeDetectorRef
```
pjmonks: This `ChangeDetectorRef` isn't used anywhere so can be removed.

src/app/catalogue-search/catalogue-search-profile-filter-list/catalogue-search-profile-filter-list.component.ts
```diff
@@ -121,6 +131,65 @@ export class CatalogueSearchProfileFilterListComponent
...
 
+  mapFromProfileFilters(
+    profileFilters: CatalogueSearchProfileFilter[] | undefined
+  ): FormArray {
+    const filters = new FormArray([]);
+    if (!profileFilters) {
+      return filters;
+    }
+    profileFilters.forEach((filter: CatalogueSearchProfileFilter) => {
+      const row = new FormGroup({
```
pjmonks: I think this setup of initial profile filters should probably be using the existing `createFilter()` function (with some possible refactoring). Almost the same setup is being done, and I've noticed a bug with this current implementation:

1. Start with one existing profile filter
2. Open the dialog
3. Now change the profile column
4. Key/Value are reset but the profile definition for Key field is not refreshed - the old column names still exist

Using the `createFilter()` function would fix this because the form field subscriptions are setup correctly to handle this.

You will need to adjust it though to set initial values (or not in the case of creating a new row).

src/app/catalogue-search/catalogue-search-profile-filter-list/catalogue-search-profile-filter-list.component.ts
```diff
@@ -121,6 +131,65 @@ export class CatalogueSearchProfileFilterListComponent
...
+      row.controls.key.valueChanges
+        .pipe(takeUntil(this.unsubscribe$))
+        .subscribe(() => {
+          row.controls.value.reset();
+        });
+      this.filters.push(row);
+    });
+  }
+
+  setInitialDefinitionValue(row: FormGroup, provider: ProfileSummary): void {
```
pjmonks, line 132: Combined with my comment above, this could be refactored to handle both initial setup and creating new rows. I think this would return an `Observable<ProfileDefinition>` instead of subscribing though so other function could chain it together.

src/app/catalogue-search/profile-filter-card/profile-filter-card.component.html
```diff
@@ -0,0 +1,40 @@
...
+-->
+<mat-card appearance="outline" class="profile-filter-card">
+  <button
+    class="profile-filter-card__delete-button"
+    mat-stroked-button
+    color="primary"
+    type="button"
+    (click)="onDelete()"
+  >
+    <i class="fa fa-times"></i>
```
pjmonks: Change this font awesome icon to a `<span>` to avoid later code warnings from Sonarqube, using the `<i>` tag is apparently not semantically correct.

src/app/catalogue-search/profile-filter-card/profile-filter-card.component.html
```diff
@@ -0,0 +1,40 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+SPDX-License-Identifier: Apache-2.0
+-->
+<mat-card appearance="outline" class="profile-filter-card">
+  <button
```
pjmonks, line 23: Buttons without any text should include `aria-label` for accessibility to describe the action. Also a `matTooltip` would be useful here.

src/app/catalogue-search/profile-filter-card/profile-filter-card.component.html
```diff
@@ -0,0 +1,40 @@
...
+    class="profile-filter-card__delete-button"
+    mat-stroked-button
+    color="primary"
+    type="button"
+    (click)="onDelete()"
+  >
+    <i class="fa fa-times"></i>
+  </button>
+  <div class="profile-filter-card__profile">
+    <mat-label>Profile:</mat-label>
```
pjmonks: Having the `<mat-label>` elements feels like it adds too much content to the cards. I'd suggest:

1. Removing the labels, just keep the values
2. Use `<h4>` for the provider name
3. Use `<strong>` for the key name

src/app/catalogue-search/profile-filter-card/profile-filter-card.component.html
```diff
@@ -0,0 +1,40 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+SPDX-License-Identifier: Apache-2.0
+-->
+<mat-card appearance="outline" class="profile-filter-card">
```
pjmonks: Some margin between cards would make them stand out better. You could updated `profile-filter-card` to have `margin: 8px 0` for that.

src/app/catalogue-search/profile-filter-dialog-component/profile-filter-dialog-component.component.html
```diff
@@ -0,0 +1,26 @@
...
+limitations under the License.
+
+SPDX-License-Identifier: Apache-2.0
+-->
+<mat-dialog-content>
+  <mdm-catalogue-search-profile-filter-list
+    [prefilledFilters]="data.profileFilters"
+  ></mdm-catalogue-search-profile-filter-list>
+</mat-dialog-content>
+<mat-dialog-actions>
```
pjmonks: Add `align="end"` to `<mat-dialog-actions>` to consistently align the buttons to the right.

src/app/catalogue-search/profile-filter-dialog-component/profile-filter-dialog-component.component.html
```diff
@@ -0,0 +1,26 @@
...
+  <button mat-button (click)="onCancelClick()">Cancel</button>
+  <button mat-button (click)="onSaveClick()">Save</button>
```
pjmonks: The Save button should use `mat-flat-button` and `color="primary"` to make it stand out as the primary action.

src/app/catalogue-search/profile-filter-dialog-component/profile-filter-dialog-component.ts
```diff
@@ -0,0 +1,44 @@
...
+  templateUrl: './profile-filter-dialog-component.component.html',
+  styleUrls: ['./profile-filter-dialog-component.component.scss']
+})
+export class ProfileFilterDialogComponent {
+  @ViewChild(CatalogueSearchProfileFilterListComponent)
+  profileFiltersForm: CatalogueSearchProfileFilterListComponent;
+
+  constructor(
+    public dialogRef: MatDialogRef<ProfileFilterDialogComponent>,
+    @Inject(MAT_DIALOG_DATA) public data: { profileFilters: any }
```
pjmonks: A type should be added to `profileFilters`, I think this is `CatalogueSearchProfileFilter[]`

src/app/catalogue-search/profile-filters/profile-filters.component.html
```diff
@@ -0,0 +1,53 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+SPDX-License-Identifier: Apache-2.0
+-->
+<div class="mdm-profile-filters">
+  <div class="mdm-profile-filters__header">
+    <h3>MetaData Filters</h3>
```
pjmonks: Rename the heading to "Profile Filters" to make it consistent with the main Search page.

src/app/catalogue-search/profile-filters/profile-filters.component.html
```diff
@@ -0,0 +1,53 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+SPDX-License-Identifier: Apache-2.0
+-->
+<div class="mdm-profile-filters">
```
pjmonks, line 18: Add a `margin-top: 8px` to the `mdm-profile-filters` class to add some spacing between elements in the filters column.

pjmonks (changes requested):
Mostly looking good. I have identified some cleanup tasks which I would like completed before merging, and left some suggestions on how to fix them so they can hopefully be done quickly.

src/app/catalogue-search/catalogue-search-profile-filter-list/catalogue-search-profile-filter-list.component.ts
```diff
@@ -121,23 +127,68 @@ export class CatalogueSearchProfileFilterListComponent
...
+    return (
+      a.metadataPropertyName === b.metadataPropertyName &&
+      a.description === b.description
+    );
+  }
+
+  private createFilter(filter?: FormGroup) {
+    /* eslint-disable @typescript-eslint/unbound-method */
+    if (!filter) {
+      filter = new FormGroup({
```
pjmonks: I'm afraid this change hasn't achieved what I was intending. There are 2 problems here:

1. Creating a `FormGroup` for each row is now duplicated in two places in the code, this should only be created in one place for better code maintenance.
2. The copy in `createFilter()` sets up validators for each field, but the one in `mapFromProfileFilters()` does not, already showing a divergence in functionality.

`createFilter()` should actually be accepting the `filter?: CatalogueSearchProfileFilter` as an optional parameter, then you can set the form fields during creation like this:

```ts
provider: new FormControl<ProfileSummary>(filter?.provider, Validators.required)
```

Using the null coalescing operator ensures either the value is set or `null` is used, satisfying both setup conditions.

If you need to do something further with the row created, have `createFilter()` return it - it can always be ignored if not required.

src/app/catalogue-search/profile-filters/profile-filters.component.html
pjmonks: This hasn't been actioned.

pjmonks (changes requested):
Two final things to check over please, then I'd say it's ready to go

pjmonks (approved):
Good effort 👍 I'm happy with this now.

PR #810, opened 2023-05-23
Make "Compare" and "Merge Graph" available to any logged-in user

### LINKED ISSUE #809: "Compare" option on data models is only available to administrators
When logged in as an administrator, you can compare 2 models by using the menu option in the "..." dropdown.  When logged in as a non-admin, then this option is not available.  Comparison should be available to all users.

jamesrwelch:
Fixes #809 by making "Compare" and "Merge Graph" available to any logged-in user.


joe-crawford (approved):
Great, have tested locally and all looks good :+1:

PR #812, opened 2023-06-08
Add scrollable tree view and styling changes

### LINKED ISSUE #806: Separate scrolling experience
# Description

It would be helpful to have a scroll bar for the tree of models and a separate one for the details page, rather than a single scroll bar for both.  In anger, its a pain to have to keep scrolling to see the details of the thing you select.

jamesrwelch:
Fixes #806, and as a result, also changes the page structure to include a fixed-height view with fixed header / footer.  
- Scrollable tree in both directions
- Separate scrolling for left / right panels in browse
- Reduce the height of the footer
- Fixed header / footer
- Styling to make all of the above work on all pages

Screenshot 2023-06-08 at 14 02 58


src/style/_default.scss
```diff
@@ -25,8 +25,21 @@ SPDX-License-Identifier: Apache-2.0
 body {
   height: 100vh;
   width: 100vw;
+  background-color: #eee;
```
joe-crawford: I think this could be `var(--light)` to re-use the value from the Material theme.

joe-crawford:
When viewing the front page in a narrow window, the two boxes don't have the same margins, also the image in the top box can overflow slightly.

Screenshot 2023-06-08 at 15 06 21

src/app/home/home.component.html
```diff
@@ -28,11 +35,13 @@
           <button mat-flat-button color="primary" (click)="login()" class="custom mb-1">
             Log in
           </button>
+          -->
           <!-- <span> or </span>
                     <button mat-stroked-button color="primary" (click)="register()" class="custom mb-1">
                         Register an account
                     </button> -->
-          <a mat-stroked-button color="primary" class="custom no-underline ml-1" target="_blank" [href]="documentationUrl">Learn more</a>
```
joe-crawford, line 35: Log in and learn more buttons on centre of the front page have been commented out.

joe-crawford:
In a narrow window in the 'Browse' view, the 'Hide Models' option still leaves an empty element for the tree visible, especially in a very narrow window which triggers the vertical view:

Screenshot 2023-06-08 at 15 11 25

joe-crawford:
Comments comparing old (top) with new changes (bottom):

* Additional spacing around the model component makes the spacing on the model header panel look slightly wide.
* Removal of Material shadow under model header panel makes the existing spacing look bigger.
* (not in screenshot) Spacing on the 'Default profile' dropdown has been removed to make it match static content below, but the 'Edit' button to the right still has more spacing than the static content. IMHO I can understand the interactive buttons/dropdowns having different spacing to the other content, but should be consistent.

Screenshot 2023-06-08 at 15 15 54

joe-crawford:
I've tried using the scrollable tree pane to browse a large folder of DataModels, which worked well 🎉

src/style/_default.scss
jamesrwelch: Resolved with 5bc55af88f38481e3d52f02de97cbcfa7694c7c0

src/app/home/home.component.html
jamesrwelch: Both these issues fixed with 6bff91868e50914faa09a5f037f05334b6db35be

jamesrwelch:
Also fixed the "Hide Models" issues with 1392338dedad1636164bd416c750b087ba6e6967

jamesrwelch:
Commit 6140ec4419b184601c063eb337fec0e2b785bc27 fixes the alignment on the model details header, and the edit button.  Oddly (maybe), it's more pleasing to align the text with the text in the table below, rather than the edge of the table.
There may be other alignment issues, but I'll be happy to work through those systematically as another ticket.  
Ready for re-review - thanks for the comments so far :)

joe-crawford (approved):
The updates look good to me :+1:

Noted that there are still various alignment issues to be handled later.

PR #813, opened 2023-06-12
Merge Data Classes and Data Elements components

jamesrwelch:
This feature will keep Jim happy.  In the Data Class page, merges the Data Class and Data Element lists into a single component.  This means we can hide one of those tables when the contents are empty, and move the "Add..." button to the other table if the parent class is editable.  
Other minor fixes:
- Removes the "skeleton" table that acts as a placeholder while the contents are loading.  Because this is never the same height as the resulting table, it causes jerkiness when there are two tables on the page.
- Changes the tab name from "Elements" to "Contents" because I think that's clearer
- Removes some other unnecessary code from the TS definitions.
- Adds a new `div` for when both tables are empty.


joe-crawford (changes requested):
Some general comments after testing the UI:

Testing in Chrome, when typing into the filter boxes, sometimes the input loses focus on the box so further characters can't be typed. After clicking on the filter box again, it loses focus again after each extra character. In the screenshot below this happened on the Data Classes name filter:
Screenshot 2023-06-12 at 16 26 02

Two 'Add Data Class' buttons appeared after applying a filter:
Screenshot 2023-06-12 at 16 29 52

After applying a filter on a Class which only had Data Elements contents and no Data Classes, the no contents message was duplicated:
Screenshot 2023-06-12 at 16 28 26

jamesrwelch:
Thanks for your thorough testing Joe - definitely a few logical errors in there that I hadn't spotted!  I think this is ready for re-testing now.

joe-crawford (approved):
I've retested and this all looks good to me :+1: Happy to merge :)

PR #815, opened 2023-06-15
GH-416 Add show superceded document search filter

GBishop-PHSA:
gh-416 Addition of a search filter to filter out superceded document versions


joe-crawford (changes requested):
Thanks for this, I've tested locally and this looks good to me.

As in the core change, can I suggest the name of this is changed to just `Show superseded` or `Show superseded models` as it is using both the model version and the document version (which is correct) and we are also trying to move away from using the document version.

GBishop-PHSA:
@joe-crawford I've updated the variable name to be consistent with mdm-core.

joe-crawford (approved):
Great, thanks for this :+1:

PR #816, opened 2023-06-15
GH-415 Config to include comments and files during catalogue copy

GBishop-PHSA:
gh-415 New configuration to include commnets and files during a catalogue copy action for new versions


joe-crawford (approved):
This looks good to me, thanks. I will wait on https://github.com/MauroDataMapper/mdm-core/pull/425.

joe-crawford:
Sorry for the delay in merging this! We should probably enable this by default anyway.

Issue #808, opened 2023-05-19
Add file size validation to the Annotations > Add Attachment screen


GBishop-PHSA:
# Description

When user's add a file attachment, the UI does not validate the size of the file. Our administrators would like the option to limit the file size and provide an error message to the user if a selected file exceeds the limit.

# Additional context

Add a configuration feature.attachment_size_limit_mb, where the value is a number representing the file size limit in megabytes.
The user will get an error response from the UI if the limit is exceeded. 

The save method of the attachment-list component will include a check for file size. If a file limit config is set and the file exceeds that limit, an error message, "There was a problem saving the attachment. Files cannot be larger than XXmb"

joe-crawford:
Hi @GBishop-PHSA,

This sounds like it would be a useful feature. Just FYI there's also a Tomcat property that is set when running in the Docker container which may be relevant, the name is `MAX_UPLOAD_SIZE`, there's also a grails.controllers.upload.maxFileSize param in mdm-core.

GBishop-PHSA:
@joe-crawford Since max size is set in the core app and tomcat, is it not possible to have a configurable value to use for validation by the UI? Should core return its set max to the UI for validation instead?

joe-crawford:
At the moment I don't think there are any ApiProperties that are fully dynamic and controlled by the backend, although I don't think there would be a problem with adding one if it's readonly.

Alternatively it could be added as a normal configurable ApiProperty but you could use BootStrap.groovy to set the appropriate value on first startup, hopefully there is some way of reading the tomcat configuration value in there as well as the Grails one. I wouldn't have an objection to that but when changing the limit we'd have to remember to change it there as well.

GBishop-PHSA:
I like the idea of setting the value for ApiProperty during bootstrap, so that the UI can read it. Is there a current way to set an ApiProperty to be read-only? I wouldn't want an admin to change the value since it wouldn't match the mdm-core/tomcat values.

joe-crawford:
:+1: Maybe BootStrap.groovy could reset the correct value on every start up. I don't think there's currently a read-only property, to add one you'd have to add a Boolean flag on ApiProperty, and add a SQL migration script to add the corresponding column to the DB.

GBishop-PHSA:
@jamesrwelch I've updated mdm-core to validate that an admin has set a max file upload size no greater than the maximum set in the Grails plugin.yml. Bootstrap was update to create an initial value for the api property.
[...]

PR #826, opened 2023-07-24
gh-808 Attachment file size limit validation

GBishop-PHSA:
gh-808 When users upload a file that exceeds the mazimum file size limit, the will receive a helpful error message based on a value set in the admin configuration.


joe-crawford:
Hi @GBishop-PHSA, thanks for this contribution. I'm running `feature/gh-808` of core and the UI but I haven't managed to see the message when uploading a large file.

I can see errors in the console when loading the Annotations tab, one is
```
ERROR SyntaxError: "undefined" is not valid JSON
    at JSON.parse (<anonymous>)
    at AttachmentListComponent.loadAttachmentFileSizeLimit (attachment-list.component.ts:317:41)
```

However when uploading a larger file (170MB so shouldn't be over the limit) I'm also getting another error:
```
ERROR RangeError: Invalid array length
    at Function.from (<anonymous>)
    at reader.onload (attachment-list.component.ts:286:29)
    at _ZoneDelegate.invoke (zone.js:375:26)
    at Object.onInvoke (core.mjs:26274:33)
    at _ZoneDelegate.invoke (zone.js:374:52)
    at Zone.runGuarded (zone.js:144:47)
    at FileReader.<anonymous> (zone.js:128:29)
```

Is this something that you've seen? 

I'm using Google Chrome 115 on macOS.

GBishop-PHSA:
@joe-crawford I've fixed the json parsing exception. I wasn't able to repeat the array length error. I think that code change was added from May 2021.

joe-crawford:
Thanks for the update @GBishop-PHSA, I'm trying this again on your branch of mdm-core and mdm-ui, I'm trying to test with the default value of 200MB set in mdm-core, when trying to upload a large file again I'm seeing the same `Invalid array length` error as last time. I'm on latest Chrome on macOS. Do you have any suggestions for how to test this locally?

joe-crawford:
Screenshot 2024-04-04 at 16 12 10
Each time I click the green tick to upload the file, the Invalid array length error repeats, it takes about a second after clicking to happen.

File size is 210MB so was hoping it would trigger the message.

joe-crawford:
I think the issue may be with the existing
```fileContents: Array.from(fileBytes)```
as looking at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Invalid_array_length:
>For [Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array) the maximum length is 2^32-1

and perhaps the Chrome limit is smaller than this? 

Would it be possible to fix this, or if not in this issue, at least adjust the default down so it's less than the upper array limit size on Chrome?

joe-crawford (approved):
Thanks for this and sorry for the delay in merging!

Using Firefox on macOS I've been able to see the warning successfully when trying to upload an oversized file with the default limit (190MB). As mentioned Chrome on macOS (arm64) fails on smaller files due to the array issue. A workaround for now could be lowering the default limit to something that works on macOS Chrome. Will leave that for another PR.

Issue #833, opened 2023-11-03
Error on link suggestion


jamesrwelch:
When getting link suggestions for an individual data element, the UI hangs while processing the response from the API.
The console error is:
```
TypeError: this.model.suggestions.forEach is not a function. (In 'this.model.suggestions.forEach((suggestion) => { suggestion.selectedTarget = suggestion.results[0]; })', 'this.model.suggestions.forEach' is undefined)
```
Reported by Thomas Williamson via Zulip

PR #834, opened 2023-11-03
Fix for #833 - fix array when suggestions are for a single data element

jamesrwelch:
Minor fix for the link suggestions page


joe-crawford:
Great, this works! Thanks.

joe-crawford (approved):

PR #862, opened 2024-05-09
gh-860 Intercept HTML content to display correctly in the html-editor-component

### LINKED ISSUE #860: Rewrite internal Mauro path URLs for HTML content
# Description

The Markdown editor has a custom renderer for the final markdown content to display, where links in anchor tags are rewritten as they are found so that they point to the correct `mdm-ui` pages. The links will be stored in this markdown format:

```md
[My data model](dm:My data model|dc:My data class)
```

This obviously produces an invalid URL for the `href` attribute, but when rewritten it will produce something like this:

```html
<a href="#/catalogue/items/dataModels/dm:My data model|dc:My data class">My data model</a>
```

The HTML editor component should be updated to do something similar - allow an internal Mauro path to be used as a link, but when the content is rendered it should automatically put the correct URL prefix before that.

pjmonks:
Resolves #860 

- Add a HtmlParserService to modify links in content
- Adjust any link hrefs that contain just a Mauro path to use a correct Angular routing URL to a page
- Fix inserting links in HTML/Markdown editors. Use the correct path information for inserting links from either autocomplete or dialogs

This should work in conjunction with the issue below, so that HTML content is stored maintaining Mauro paths as hyperlinks, yet will correct them at display time:

- MauroDataMapper-NHSD/mdm-plugin-nhs-data-dictionary#11

### DIFF
--- a/src/app/content/content.utils.ts
+++ b/src/app/content/content.utils.ts
@@ -1,4 +1,4 @@
-<!--
+/*
 Copyright 2020-2024 University of Oxford and NHS England
 
 Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


jamesrwelch (approved):
This works nicely, and the code looks sensible.  Loving the tests!

PR #866, opened 2024-05-31
gh-865: Allow localhost in Urls

### LINKED ISSUE #865: Allow localhost in Urls
# Description

`localhost` is banned from all URL strings. This needs to be relaxed to allow us to develop and test against local running applications.

# Additional context

`mdm-validators` has a complicated regex to enforce this rule. This needs to be updated to allow `localhost` as well.

NigelPalmer:
Closes #865 

This can be tested by creating an Open ID Connect configuration and typing in the Discovery Url field. Before this change is applied, localhost is not allowed. Afterwards it is.


jamesrwelch (approved):
Nice one, @NigelPalmer , thank you :)

PR #871, opened 2024-06-20
gh-870, make definition mandatory in line with other add terminology …

### LINKED ISSUE #870: Creating/editing Terms must include a "Definition" despite appearing to be an optional field
# Description

The "Create New Term" dialog appears to state that the "Definition" field is optional (unlike "Code"), yet when trying to save a term without a definition, it gives this error:

Image

# Steps to reproduce

1. Go to one of the Terminologies
2. Add a new term
3. Enter a code at least, leave the other fields blank
4. Click "Create"

# Expected behavior

Either the definition field should be mandatory, or it can remain optional and automatically copy the chosen code to the definition to provide a value before sending to the backend.

# Related

- NHSDigital/DataDictionaryPublication#474
- NHSDigital/DataDictionaryPublication#473
- NHSDigital/DataDictionaryPublication#472

OButlerOcc:
…screens

I had a look at the terminology screen for adding a new term when creating a link and decided since that one has the field as required, I would follow suite. 

Closes #870

### DIFF
--- a/src/app/terminology/term-list/create-term-dialog/create-term-dialog.component.html
+++ b/src/app/terminology/term-list/create-term-dialog/create-term-dialog.component.html
@@ -31,9 +31,9 @@ SPDX-License-Identifier: Apache-2.0
       </mat-form-field>
     </div>
     <div class="mdm--form-input">
-      <mat-form-field appearance="outline" class="full-width">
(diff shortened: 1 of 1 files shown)


pjmonks (approved):
This is good. It's an efficient change and has clearly been tested, I've confirmed this works too. Approved 👍

PR #875, opened 2024-07-03
gh-874 Create term dialog sets default definition

### LINKED ISSUE #874: Creating/editing a term needs to auto-fill definition if not provided
# Description

This is related to the previous issue:

- #870 

NHSE still think that having a mandator "definition" field for a term is not acceptable. Instead, the field should be made optional but auto filled before submitting with the code if not provided.

# Related

- NHSDigital/DataDictionaryPublication#474
- NHSDigital/DataDictionaryPublication#473
- NHSDigital/DataDictionaryPublication#472

pjmonks:
If no definition provided, it will auto default to the same value as the code. Definition field is now optional again.

Resolves #874 

To test:

1. Select a terminology that can be edited
2. Go to the "Terms" tab and click "Add term"
3. Enter a code but no definition, then click "Create". This should produce a term where code and definition are the same.
4. Repeat but add a code and definition that are different values. This should produce a term with the values you have set.

### DIFF
--- a/src/app/terminology/term-list/create-term-dialog/create-term-dialog.component.html
+++ b/src/app/terminology/term-list/create-term-dialog/create-term-dialog.component.html
@@ -27,13 +27,15 @@ SPDX-License-Identifier: Apache-2.0
       <mat-form-field appearance="outline" class="full-width">
         <mat-label>Code</mat-label>
         <input matInput type="text" formControlName="code" required />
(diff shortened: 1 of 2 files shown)


OButlerOcc (approved):
Tests fine, populates based on term if nothing is set,
definition field is not mandatory

PR #884, opened 2024-07-16
gh-883 Change to support history tracking

### LINKED ISSUE #877: History tab missing on DataElements
# History tab missing

The User Interface for Data Elements does not enable the user to see the history of change for the data element.  The parent class history does not show the changes to elements on the classes either

# Suggested resolution

Implement a means to view the history

### LINKED ISSUE #883: Changes to support history tracking of items
# Description

Make the following changes to support tracking of history for Mauro items:

- [x] Add a column to the "History" tab called "Action" - this should be the returned `title` property in each row.
- [x] Make sure the "History" tab is visible in every relevant item domain type.

pjmonks:
- Add an "Action" column to the History table to determine the action that was carried out e.g. "CREATE", "COPY", "UPDATE" etc
- Add the History tab to missing domain types - Data Element, Term and Data Type

image

Resolves #883 
Resolves #877

### DIFF
--- a/src/app/data-type/data-type.component.html
+++ b/src/app/data-type/data-type.component.html
@@ -108,5 +108,21 @@ SPDX-License-Identifier: Apache-2.0
         </div>
       </ng-template>
     </mat-tab>
+    <mat-tab>
+      <ng-template mat-tab-label>
+        History
+        <mdm-skeleton-badge *ngIf="isLoadingHistory"></mdm-skeleton-badge>
+        ({{ historyItemCount }})
+      </ng-template>
+
+      <div class="full-width">
+        <mdm-history
+          (totalCount)="historyCountEmitter($event)"
+          *ngIf="dataType"
+          [parent]="dataType"
+          [domainType]="'dataTypes'"
+        ></mdm-history>
+      </div>
+    </mat-tab>
   </mat-tab-group>
 </div>
--- a/src/app/data-type/data-type.component.ts
+++ b/src/app/data-type/data-type.component.ts
@@ -62,6 +62,8 @@ export class DataTypeComponent
   contextView = 'default';
   rulesItemCount = 0;
   isLoadingRules = true;
+  historyItemCount = 0;
+  isLoadingHistory = true;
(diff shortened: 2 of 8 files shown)


OButlerOcc (dismissed):
Looks fine to me, tested fine, wow the formatting on that before you got to it was weird

mhocc (dismissed):
This all seems good, the Action column is in place and history appears for the additional domain types.

Separately I noticed there seems to be a limit of 10 history items if updates are made after 10 changes then nothing gets added to the history.  It does not appear to be possible to sort by Date — clicking on the column header has no effect.  Also, in some places the number of history items does not appear next to the History tab, for example Data Sets and Data elements do display the count while Data classes do not.

pjmonks:
@mhocc I had to push a change so that Jenkins would trigger a new build, so I looked at the sorting/pagination issue you raised and fixed that. Sorting by date (or another field) does work now, pagination I assume to work but haven't got enough history rows to confirm. The `GET` request sent is at least sending the correct query string now though. Please could you review again to confirm.

mhocc (commented):
I can see the sorting works, the UI now supports 50 items per history page but if I add more than 50 it continues to report 50 and the paging controls remain disabled.  The GET command sends `max=50` and `sort` parameters, the response returns a count of items (`count: 50`) but I've not seen anywhere that return the total number of items so it looks like paging is not supported by the backend.


Found a a minor issue—clicking the date column header first sorts in ascending order, another click sorts in descending order and a third click *seems* to sort in ascending order but there's no arrow (the GET does not send a `sort` parameter), a fourth click keeps the same order but adds an arrow.

pjmonks:
@mhocc Based on your last comment, could you please:

1. Create a new issue under `mdm-ui` to explain the pagination issues you're seeing.
2. Approve this PR as it is, since it builds and _mostly_ works. I can accept some known issues at this point, we'll just progress forward.

Once this PR is approved I'm able to merge and close it.

mhocc (approved):
Approved

Created #890 as a follow-on task

PR #888, opened 2024-07-17
gh-886 HtmlParserService now generates Mauro path URLs relative to a version or branch

### LINKED ISSUE #886: Allow item lookup from local paths
At the moment items can be looked up by path - e.g. `dm:Test data model$1.0.0|dc:Test data class`.  This is used in the markdown descriptions of items and works nicely when referring to a finalised item elsewhere in the system.
However, if you're referring to something within the same model, or the same versioned folder, which is not yet finalised, then this causes problems - e.g. all items of `dm:Test data model$main|dc:Test data class` when the `Test data model` was finalised, or the branch name was changed.

A 'local' path notation is required so that we can refer to items within the same model / vf without specifying the branch name which is subject to change.

This is a requirement for the Data Dictionary work to save huge amounts of renaming and to generally improve the experience of using links in descriptions.

pjmonks:
Resolves #886 

# Overview

- HtmlEditor will determine the version/branch the content is relative to based on the provided root element
- If known, all paths found in the HTML source description text will be rewritten as valid URLs and patched to work relative to the version/branch selected

In the example below, notice that the current item is under a branch called "CR1234 - Test Branch". This means the hyperlinks in the description field have automatically been adjusted to also use that branch name.

image

# Testing

1. Create an NHS Data Dictionary branch called `main`
2. Create another NHS Data Dictionary branch with a different name
3. Under `main`, navigate to items with descriptions and check the hyperlinks. These links to navigate successfully to items also under the `main` branch.
4. Repeat step 3 but in the other branch - this time, hyperlinks should navigate successfully to items under the other branch name, _not_ the `main` branch

### DIFF
--- a/src/app/content/html/html-editor/html-editor.component.ts
+++ b/src/app/content/html/html-editor/html-editor.component.ts
@@ -155,7 +155,11 @@ export class HtmlEditorComponent implements OnInit, OnChanges {
       }
 
       // Intercept the HTML content to render for display only so that certain parts can be altered
-      this.displayContent = this.htmlParser.parseAndModify(this.description);
+      // The version or branch override will try to force all Mauro path links to work within the same context as
+      // the containing model the root element is in, if available
+      this.displayContent = this.htmlParser.parseAndModify(this.description, {
+        versionOrBranchOverride: this.getVersionOrBranchOverride()
+      });
     }
   }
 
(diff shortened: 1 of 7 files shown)


src/app/content/html/html-editor/html-editor.component.ts
```diff
@@ -155,7 +155,11 @@ export class HtmlEditorComponent implements OnInit, OnChanges {
       }
 
       // Intercept the HTML content to render for display only so that certain parts can be altered
-      this.displayContent = this.htmlParser.parseAndModify(this.description);
+      // The version or branch override will try to force all Mauro path links to work within the same context as
```
OButlerOcc, line 158: Comment here helps a lot

src/app/content/html/html-parser/html-parser.service.spec.ts
```diff
@@ -69,13 +82,42 @@ describe('HtmlParserService', () => {
...
     );
 
     const content = `<a href="${path}">Mauro item link</a>`;
     const expected = `<a href="${baseUrl}${pathUrl}">Mauro item link</a>`;
-    const modified = service.parseAndModify(content);
+    const modified = service.parseAndModify(content, {});
     expect(modified).toStrictEqual(expected);
   });
+
+  it.each([
```
OButlerOcc, line 95: this helps a lot understanding the context of the change. I think this is a good test

OButlerOcc (approved):
Tests fine, thankyou for including the script, I wasn't sure exactly what I was looking for. Unit tests are good, thank you for those, very clear.

jamesrwelch (approved):
This works really nicely - love it! 🥇

PR #891, opened 2024-07-19
gh-889 - History tab for types was not functioning

### LINKED ISSUE #889: History tab of data types does not display records
# Description

When viewing the "History" tab of a data type - a primitive type, model type, etc - no records appear in the table.

# Steps to reproduce

1. Create multiple changes in a primitive data type under a data model - either create a type, modify a type, merge changes from a type into another branch etc.
2. View the data type
3. Click on the "History" tab.

# Additional context

The problem I've found is the UI sends the wrong domain type in the request, it sends this:

```http
GET api/dataTypes/{id}/edits
```

When it should be more specific, like this:

```http
GET api/primitiveTypes/{id}/edits
```

Checking the Postgres database proves that the edit history is recorded for the data type, but it must be filtered by the correct resource domain type:

image

OButlerOcc:
Closes #889 

History tab of data types did not work because it was not modifying the endpoint called based on the type of dataType.
added a getter that grabs that info and forwards it to the history tab.

To test: 

1. Create changes in a variety of data type under a data model - either create a type, modify a type, merge changes from a type into another branch or add a rule. (possible future improvement, adding a rule does not refresh the history tab)
2. View the data type
3. Click on the "History" tab.
4. Note all the shiny entries

image

### DIFF
--- a/src/app/data-type/data-type.component.html
+++ b/src/app/data-type/data-type.component.html
@@ -120,7 +120,7 @@ SPDX-License-Identifier: Apache-2.0
           (totalCount)="historyCountEmitter($event)"
           *ngIf="dataType"
           [parent]="dataType"
-          [domainType]="'dataTypes'"
+          [domainType]="typeOfDataType"
         ></mdm-history>
       </div>
     </mat-tab>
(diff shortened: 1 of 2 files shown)


pjmonks (approved):
Works great 👍

PR #893, opened 2024-07-23
Feature/gh 890   pagination in history tab

### LINKED ISSUE #890: Fix pagination in history tracking
# Description

There is history tracking tab on the UI for a number of item types (data classes, folders, data standards, data elements etc.) when the number of items in the list exceed the page size (currently set to 50) it should be possible to page through the list using the controls at the bottom of the page—currently this does not work.

In addition, the total number of history items appear next to the label on the history tab for some types (e.g. `DataElement`) but not others (`DataClass`) this behaviour needs to be consistent.

Toggling the date sorting by clicking on the "**Date**" column header cycles through: Ascending → Descending → (no sorting) → Ascending.

# Steps to reproduce

1. Navigate to an item in the UI and open the "**History**" tab
2. If there are less than 50 items in the list make edits (to the description, label etc.) until there are more than 50 item in the list
3. Notice that the total number displayed on the page is 50, next scroll to the bottom of the page and notice the pagination controls are disabled
4. Sort the list by date and notice that there are indeed more than 50 items

# Expected behaviour

* When there are more then 50 items in the history list it should be possible to page through the items while retaining the current sort order
* The total number of history items either needs to be displayed for all history types or none.

OButlerOcc:
Closes #890

Requires backend change: https://github.com/MauroDataMapper/mdm-core/pull/484

History tab was not correctly receiving the totalCount from the backend, bricking the paginator.
while I was fixing the backend I ended up changing a few things to make the code more obvious, I'd like to keep the changes

To test:
Get application build or mdm core up and running with the new backend

1. Navigate to an item in the UI and open the "History" tab
2. If there are less than 50 items in the list make edits (to the description, label etc.) until there are more than 51 items or more in the list
3. Notice that the total number displayed on the page is correct, next scroll to the bottom of the page and notice the pagination controls are working correctly
5. Rejoice (optional)

### DIFF
--- a/src/app/shared/history/history.component.html
+++ b/src/app/shared/history/history.component.html
@@ -171,13 +171,14 @@ SPDX-License-Identifier: Apache-2.0
   </table>
 </div>
 <div
-class="mdm--mat-pagination"
-[ngClass]="{ 'is-hidden': totalItemCount < 6 }"
+  class="mdm--mat-pagination"
+  [ngClass]="{ 'is-hidden': totalItemCount < 6 }"
 >
-<mdm-paginator
-  [length]="totalItemCount"
-  [pageSize]="pageSize"
-  pageIndex="0"
-  showFirstLastButtons
-></mdm-paginator>
+  <mdm-paginator
+    [length]="totalItemCount"
+    [pageSize]="pageSize"
+    [pageSizeOptions]="pageSizeOptions"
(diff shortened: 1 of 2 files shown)


pjmonks (approved):
I simply added a page size options list to the history table, which was missing anyway. I've tested with the backend change and it works great now.

PR #895, opened 2024-07-24
Fixes for the Element Selector dialog and link creation in HTML editor

### LINKED ISSUE #882: Model tree control does not show Versioned Folders
# Description

When using the Model Tree control is various parts of the UI, you cannot expand folders to display child versioned folders, excluding a number of models beneath them.

# Steps to reproduce

Edit any description, then switch to the "HTML" format.

Click on this link button:

Image

Select an option like "Data Model", or "Data Class" to show the model tree. Then expand to where you think a versioned folder should appear as a child item in the tree:

Image

In the above screenshot, there should be child items for this folder (which are returned by the backend), yet they do not appear in the tree control.

# Expected behavior

Versioned folders should appear in the model tree to expand and/or select.

# Additional context

Check other areas where selecting from a model tree is required e.g. global search filters.

Also check that you can add any link to the HTML text area via the dialog picker.

# Related

- NHSDigital/DataDictionaryPublication#388
- NHSDigital/DataDictionaryPublication#377

pjmonks:
Resolves #882 

# Overview

Fixed the element selector dialog when using the HTML editor, triggered by this button:

image

All element types can now expand the model trees so that versioned folders are included:

image

I've updated the Terms selection to work without loading every Terminology in the catalogue, restrict to a context:

image

And I've updated the creation of the hyperlink when added to the HTML editor to remove the branch name from the path in the link. This is because all links should be considered local to the branch they are directly under, other link rewriting code will automatically fill in a branch so the links work correctly, but they must be saved to Mauro without branches in them.

# Testing

1. Test the Element Selector - make sure that you can create a link of every type
2. Test that the links created in the HTML editor work and are constrained to the branch they are under, even if you select an element from another versioned folder branch.

### DIFF
--- a/src/app/content/html/html-editor/html-editor.component.ts
+++ b/src/app/content/html/html-editor/html-editor.component.ts
@@ -234,11 +234,19 @@ export class HtmlEditorComponent implements OnInit, OnChanges {
     const path =
       element.path ?? component.pathNames.createFromBreadcrumbs(element);
 
+    // Need to do this for NHS England, local paths should be used so that links can be maintained within the context of the branches
+    // they live under
+    const pathElements = this.pathNames.parse(path);
+    const pathElementsWithoutBranch = this.pathNames.removeVersionOrBranchName(
+      pathElements
+    );
+    const pathWithoutBranch = this.pathNames.build(pathElementsWithoutBranch);
+
     // The href will be the Mauro item path. This is not a valid URL, but the HtmlParserService will
     // automatically look for these paths in the content and rewrite them to correct URLs before
     // rendering for display
     const html = editor.create.fromHTML(
-      `<a href='${path}' title='${element.label}'>${element.label}</a>`
+      `<a href='${pathWithoutBranch}' title='${element.label}'>${element.label}</a>`
     );
 
     editor.selection.setCursorIn(focusNode);
--- a/src/app/shared/path-name/path-name.service.spec.ts
+++ b/src/app/shared/path-name/path-name.service.spec.ts
@@ -612,4 +612,26 @@ describe('PathNameService', () => {
       }
     );
   });
+
+  describe('remove branch name', () => {
+    it.each([
+      ['dm:Data Model$main', 'dm:Data Model'],
+      ['dm:Data Model$main|dc:Data Class', 'dm:Data Model|dc:Data Class'],
+      ['dm:Data Model$another', 'dm:Data Model'],
+      [
+        'dm:Data Model$another|dc:Data Class|de:Data Element',
+        'dm:Data Model|dc:Data Class|de:Data Element'
+      ]
+    ])(
+      'should remove the branch name from %p to give %p',
+      (originalPath, expectedPath) => {
+        const originaPathElements = service.parse(originalPath);
+        const updatedPathElements = service.removeVersionOrBranchName(
+          originaPathElements
+        );
+        const actualPath = service.build(updatedPathElements);
+        expect(actualPath).toBe(expectedPath);
+      }
+    );
+  });
 });
--- a/src/app/utility/element-selector.component.html
+++ b/src/app/utility/element-selector.component.html
@@ -139,8 +139,8 @@ SPDX-License-Identifier: Apache-2.0
               [inSearchMode]="formData.inSearchMode"
               [doNotMakeSelectedBold]="true"
               [searchCriteria]="formData.treeSearchText"
-              [filterByDomainType]="['Folder', 'CodeSet']"
-              [expandOnNodeClickFor]="['Folder']"
+              [filterByDomainType]="['Folder', 'VersionedFolder', 'CodeSet']"
+              [expandOnNodeClickFor]="['Folder', 'VersionedFolder']"
               (nodeClickEvent)="onNodeInTreeSelect($event)"
               [node]="rootNode"
             ></mdm-folders-tree>
@@ -167,8 +167,8 @@ SPDX-License-Identifier: Apache-2.0
               [inSearchMode]="formData.inSearchMode"
               [doNotMakeSelectedBold]="true"
               [searchCriteria]="formData.treeSearchText"
-              [filterByDomainType]="['Folder', 'DataModel']"
-              [expandOnNodeClickFor]="['Folder']"
+              [filterByDomainType]="['Folder', 'VersionedFolder', 'DataModel']"
+              [expandOnNodeClickFor]="['Folder', 'VersionedFolder']"
               (nodeClickEvent)="onNodeInTreeSelect($event)"
               [node]="rootNode"
             ></mdm-folders-tree>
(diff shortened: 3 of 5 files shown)


OButlerOcc (changes requested):
Problems with this one I'm afraid.

Adding a term fails, clicking one crashes the editor box
Seems to break at editor.selection.setCursorIn(focusNode), like 252/html-editor.componant

message.service.ts:99 ERROR TypeError: Node element must be in editor
    at t.error (jodit.min.js:1:237589)
    at e.setCursorIn (jodit.min.js:1:272131)
    at HtmlEditorComponent.createAndInsertLink (html-editor.component.ts:252:22)
    at html-editor.component.ts:197:14
    at Object.next (Subscriber.js:110:13)
    at SafeSubscriber._next (Subscriber.js:60:26)
    at SafeSubscriber.next (Subscriber.js:31:18)
    at Subject.js:31:30
    at errorContext (errorContext.js:19:9)
    at Subject.next (Subject.js:26:21)
    
Secondly, the links don't actually work when clicked, they look like: dm:give test|dt:source_ethnicity_code. The two problems seem related. I think its crashing out before inserting the link properly. commenting the editor.selection.setCursorIn(focusNode) at 252 seems to fix the term problem and the link problem but we might want to check what its doing before then.

src/app/utility/element-selector.component.html
```diff
@@ -139,8 +139,8 @@
               [inSearchMode]="formData.inSearchMode"
               [doNotMakeSelectedBold]="true"
               [searchCriteria]="formData.treeSearchText"
-              [filterByDomainType]="['Folder', 'CodeSet']"
-              [expandOnNodeClickFor]="['Folder']"
+              [filterByDomainType]="['Folder', 'VersionedFolder', 'CodeSet']"
```
OButlerOcc, line 142: I noticed that some of these selections have empty folders as options as things, as an improvement, we could steal the ones I used for the copy screen that try and filter these out. They also do things like only show terminologies for terminology selectors and such

src/app/content/html/html-editor/html-editor.component.ts
```diff
@@ -234,11 +234,19 @@ export class HtmlEditorComponent implements OnInit, OnChanges {
...
+
     // The href will be the Mauro item path. This is not a valid URL, but the HtmlParserService will
     // automatically look for these paths in the content and rewrite them to correct URLs before
     // rendering for display
     const html = editor.create.fromHTML(
-      `<a href='${path}' title='${element.label}'>${element.label}</a>`
+      `<a href='${pathWithoutBranch}' title='${element.label}'>${element.label}</a>`
     );
 
     editor.selection.setCursorIn(focusNode);
```
OButlerOcc, line 252: this line breaks things, not sure what its trying to do, commenting it fixes the issues in my testing

src/app/shared/path-name/path-name.service.spec.ts
```diff
@@ -612,4 +612,26 @@ describe('PathNameService', () => {
       }
     );
   });
+
+  describe('remove branch name', () => {
```
OButlerOcc, line 616: Nice clean easy to follow test! take it out, it doesn't belong in mauro

src/app/content/html/html-editor/html-editor.component.ts
```diff
@@ -234,11 +234,19 @@ export class HtmlEditorComponent implements OnInit, OnChanges {
...
+    const pathElements = this.pathNames.parse(path);
+    const pathElementsWithoutBranch = this.pathNames.removeVersionOrBranchName(
+      pathElements
+    );
+    const pathWithoutBranch = this.pathNames.build(pathElementsWithoutBranch);
+
     // The href will be the Mauro item path. This is not a valid URL, but the HtmlParserService will
     // automatically look for these paths in the content and rewrite them to correct URLs before
     // rendering for display
     const html = editor.create.fromHTML(
```
OButlerOcc, line 248: the link is fine here all works, its just after it gets borked

OButlerOcc (changes requested):
Problems with this one I'm afraid.

Adding a term fails, clicking one crashes the editor box
Seems to break at editor.selection.setCursorIn(focusNode), like 252/html-editor.componant

message.service.ts:99 ERROR TypeError: Node element must be in editor
    at t.error (jodit.min.js:1:237589)
    at e.setCursorIn (jodit.min.js:1:272131)
    at HtmlEditorComponent.createAndInsertLink (html-editor.component.ts:252:22)
    at html-editor.component.ts:197:14
    at Object.next (Subscriber.js:110:13)
    at SafeSubscriber._next (Subscriber.js:60:26)
    at SafeSubscriber.next (Subscriber.js:31:18)
    at Subject.js:31:30
    at errorContext (errorContext.js:19:9)
    at Subject.next (Subject.js:26:21)
    
Secondly, the links don't actually work when clicked, they look like: dm:give test|dt:source_ethnicity_code. The two problems seem related. I think its crashing out before inserting the link properly. commenting the editor.selection.setCursorIn(focusNode) at 252 seems to fix the term problem and the link problem but we might want to check what its doing before then.

src/app/utility/element-selector.component.html
pjmonks: At a later date maybe

src/app/content/html/html-editor/html-editor.component.ts
pjmonks: Links to other items within Mauro are not regular hyperlinks, they are Mauro paths. This is obviously not a valid URL:

```
<a href="dm:Model$main|dc:Class|de:Element">Some item</a>
```

But there is other code running during the read only view of the content which turns this path into the valid URL for the UI to understand.

Mauro paths are used to be agnostic to branches, so that when they are finalised or merged into other branches the links won't break.

src/app/content/html/html-editor/html-editor.component.ts
pjmonks: The link is not meant to work whilst you're still in the editor, but when you save the changes and go back to the item detail view, the links will work because they are automatically converted.

OButlerOcc:
small update, same error as term seems to happen on data model as well

pjmonks:
@OButlerOcc I tried fixing the error you found but can't make a nice fix for it. The problem is that the cursor may not be directly inside the editing area as you insert a link. For example:

1. Edit a description of an item, like the class "ACCOMMODATION".
2. _Do not click inside the text area yet!_ ... but click on the element selector toolbar button (looks like two interlocked links)
3. Select any item to insert as a link
4. The error will happen because the edit text area does not have focus.
5. Now click in the editor text area to mark a selection for the Jodit editor and repeat the steps.
6. Now the link will be inserted correctly.

I've tried to capture the exception thrown and show a nicer warning, but the warning appears too often even when it does work. I think we just have to accept this bug for now.

As for the links and paths, the links inserted will not work inside the editor as designed because they are Mauro paths and not valid URLs. When you save them and view the read-only view of the content again, then the links work because an automatic parsing function is called to convert the Mauro paths in anchor tags into valid URLs. This is how it is intended to work.

I think if there is nothing else then this sounds ready to be accepted.

src/app/content/html/html-editor/html-editor.component.ts
OButlerOcc: ahh I see, please ignore that then. I meant that when the link is created it is not added to the description at all if the node line crashes out

OButlerOcc (approved):
Currently the error occurs when the curser is not focused on the description box, this has been deemed as acceptable for now we'll fit it later.

PR #897, opened 2024-07-25
gh-896 Fix copy screen so that it works for child data classes

### LINKED ISSUE #896: Cannot copy a Data Class which is under another Data Class
# Description

It is possible to copy a Data Class but only if it's parent is a Data Model. The UI does not fetch child Data Classes correctly.

# Steps to reproduce

1. Log in to MDM UI
2. Expand an NHS Data Dictionary branch down to "Data Sets > Clinical Data Sets > Mental Health Services Data Set"
3. Select the "PATIENT DEMOGRAPHICS" data class
4. Select "Copy" from under the triple dot menu

# Expected behavior

The copy selection screen should appear to allow the copy to work.

# Additional context

Fetching data classes from Mauro means knowing the data model ID plus an optional parent data class ID. The parent data class is missing from the fetch request, which breaks the UI and returns a "Not Found" screen.

# Related

- NHSDigital/DataDictionaryPublication#496

pjmonks:
Resolves #896 

Depends on:

- MauroDataMapper/mdm-resources#118

Changes:

- Update the URL route to provide a parent data class ID
- Improve the fetch of the source item to copy
- Use new mdm-resources endpoint
- Send correct IDs for copying the child data class

To test this works, follow the steps explained in the original issue, then also test that copying the other item types still work.

### DIFF
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -530,7 +530,7 @@ export const pageRoutes: { states: Ng2StateDeclaration[] } = {
     },
     {
       name: 'appContainer.mainApp.twoSidePanel.catalogue.dataClassCopy',
-      url: '/:domainType/:dataModelId/copy/:id/',
+      url: '/:domainType/:dataModelId/:dataClassId/copy/:id/',
       component: CopyActionComponent
     },
     {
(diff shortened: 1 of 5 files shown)


mhocc (approved):
This is working correctly; tried a number of different copyable types that worked as expected

PR #901, opened 2024-08-07
gh-898 Fix element selector dialog to use infinite scroll on data elements

### LINKED ISSUE #898: Select a Data Element for an internal link does not return all available elements
# Description

Using the element selector dialog for inserting a link in a description to another Data Element, it is not possible to select any Data Element because enough are not fetched back.

# Steps to reproduce

1. Login to the NHS England test environment
2. Select any item to edit the description
3. In the HTML editor, click on the icon to open the internal link selector dialog:

Image

4. Select "Data Element"
5. Under "Choose a Data Class", select one with a lot of Data Elements e.g. pick a data dictionary branch, select "Data Elements" section, then one of the letter classes.
6. Review the list of available Data Elements in the list below.

# Expected behavior

There may be more available Data Elements, so the list should keep loading more as you scroll down.

# Additional context

The request to list Data Elements returns a high total count but only 10 or so actual items.

If you select "Term" as a link option, that list will keep loading more results the further you scroll down:

Image

# Related

- NHSDigital/DataDictionaryPublication#377

pjmonks:
Fixes #898 

- Once a data class is selected, the data elements list is returned 40 at a time
- Scrolling down the table now correctly loads the next batch, and repeats until there are no more
- Also fixed the search input for data elements for correct filtering on data model/class context

To test, open a description in the HTML editor and select the toolbar button for adding an internal link:

image

Next select "Data Element":

image

Under "Restrict your search to a context", click on the field and select from the tree that appears a data class e.g. a dictionary branch > "Data Elements" > "A"

image

Notice that some of the results are returned:

image

Scroll down the list and then a new batch of results should appear:

image

### DIFF
--- a/src/app/utility/element-selector.component.ts
+++ b/src/app/utility/element-selector.component.ts
@@ -340,7 +340,10 @@ export class ElementSelectorComponent implements OnInit {
           context: {
             id: this.formData.currentContext.id,
             domainType: this.formData.currentContext.domainType,
-            label: this.formData.currentContext.label
+            label: this.formData.currentContext.label,
+            ...(this.formData.currentContext.domainType === 'DataClass' && {
+              dataModelId: this.formData.currentContext.modelId
+            })
           }
         })
       };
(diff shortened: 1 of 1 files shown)


mhocc (approved):
Looks good, working as required

PR #908, opened 2024-10-08
Update element link component to show version/branch of parent model

pjmonks:
Related to MauroDataMapper-NHSD/mdm-plugin-nhs-data-dictionary#144

Minor update to display the model version/branch of a parent model within a semantic link source/target. This is useful for NHSE and the Data Dictionary to help determine that links are pointing to their correct branches.

image


jamesrwelch (approved):
This looks great - thank you!

PR #912, opened 2024-11-18
Enable Micronaut OpenID login

joe-crawford:
This PR changes the behaviour triggered after going to the `/redirects/open-id-connect-redirect.html` page. When using Grails, query params are passed from the Identity Provider to the backend to handle the OpenID authentication. When using Micronaut, the login has already been handled automatically by the Micronaut provided endpoints, so if no params are provided, this PR looks up the current user by using `/catalogueUsers/currentUser`, which is added in https://github.com/MauroDataMapper/mauro-micronaut/pull/100.


jamesrwelch (approved):
This seems to work nicely - thank you!

PR #932, opened 2025-05-20
Guard for createdByUser to prevent constant looping

edwardcrichton:


jamesrwelch (approved):
I didn't find the original problem but the fix makes sense 👍

PR #936, opened 2025-06-10
Fix eslint issues

jamesrwelch:
After the upgrades to angular 17 and then to angular 18


.github/workflows/node.js.yml
```diff
@@ -51,7 +50,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: MauroDataMapper/mdm-resources
-          ref: refs/heads/develop # change this to test against another branch
+          ref: refs/heads/feature/update-node # changed this... change back to develop before merge
```
edwardcrichton: What's this mean?

src/app/about/about.component.ts
```diff
@@ -89,7 +93,7 @@ export class AboutComponent implements OnInit {
     this.resources.apiProperties
       .listPublic()
       .pipe(
-        catchError(errors => {
+        catchError((errors) => {
```
edwardcrichton, line 96: 👍

edwardcrichton (approved):

PR #952, opened 2025-07-11
fix openid connect login

joe-crawford:
Reverted a few lines to an earlier version to fix OpenID Connect login.

The way the code works doesn't really make sense, but can't log in without reverting. Suggest cleaning up as a separate PR but will have to manually test as


edwardcrichton (approved):
username versus userName in types that look like they should be compatible. All fine as a fix.

PR #964, opened 2025-12-09
Jodit updates

jamesrwelch:
- Update to latest version of Jodit
- Re-enable the source plugin to allow HTML source editing
- Use the 'Ace' editor to handle html syntax markup
- Fix an issue with the Jodit popups being obscured by the angular-material dialog box
- Fix some dark styling of the html editing box


edwardcrichton (approved):
All looks in order!

PR #966, opened 2025-12-15
Update the importable and exportable domains to include folders.

jamesrwelch:
Depends on an update to mdm-resources: https://github.com/MauroDataMapper/mdm-resources/pull/147
Aware that this won't work on the grails back end and willing to accept that for now - we can add a ticket to fix that somehow


edwardcrichton (approved):
Nice. Using folders / versioned folders is essential for data dictionaries.
Some github workflow tweaks
Adds folders and versionedFolders to import and export
Adds 2 menu items for importing folders

PR #972, opened 2026-01-08
Branch/version can come from owningModel

edwardcrichton:
A model may not 'own' the versioning information but it will point to owningModel instead; fixed unguarded use of branchName and modelVersion

element-link.component.ts uses owningModel

canChangeBranchName() returns false when versioning is owned elsewhere

Didn't find any more


edwardcrichton:
Hang on, need to rebase

jamesrwelch (approved):
Great catch, thank you!

PR #973, opened 2026-01-08
Fix the issue where the Jodit editor in HTML mode would get populated…

jamesrwelch:
… with 'undefined'.  Also fix some related console errors around missing themes and commonJS imports


edwardcrichton (approved):
Nice. I see an update to ace, a check for no content, move to async bootstrap

PR #978, opened 2026-01-26
Stop the html editor inserting empty `<p></p>` tags when the content is empty

jamesrwelch:
Attempt to fix the issue where empty paragraphs tags get inserted by the html editor


edwardcrichton (approved):
🔫 Pesky editable content.

PR #980, opened 2026-02-10
Add relative paths

edwardcrichton:
Add relative paths to production configuration (baseHref, deployUrl)

Use a proxy configuration for /api for ng serve so that:
* both production and dev retrieve ui and api in the same way from the same base url,
* and the environment.ts and environment.prod.ts both use /api rather than absolute URLS for the apiEndpoint


joe-crawford (approved):
Looks good to me!

PR #981, opened 2026-02-11
Fix issue with editing descriptions - logic issue

jamesrwelch:


edwardcrichton (approved):
✅ Yes that logic is upside down. Nice one

PR #984, opened 2026-02-12
Handle the case when resolving paths and the root Object is not set. …

jamesrwelch:
… Also try and pass the root object in as many cases as possible.


edwardcrichton (approved):
👍 Rounds up the link resolution 🤠 . Good stuff

PR #987, opened 2026-02-26
safe or {{ }} or escapeHtml

edwardcrichton:
Ensure that all [innerHTML]= use | safe: 'html'
or else are converted to {{ }}

For use with the mchighlighter, add a pipe for escaping a string for use in html


jamesrwelch (approved):
This works great!  Still more work to do on the `| safe : 'html' ` pipe, but that can follow in a separate PR.

PR #992, opened 2026-03-09
Hide the password

edwardcrichton:
A toggle button for showing or hiding passwords


jamesrwelch (approved):
Yep, brilliant, thank you - good catch and speedy fix

PR #998, opened 2026-04-13
Change regex to an equivalent without ambiguity

edwardcrichton:


jamesrwelch (approved):
Validators.pattern(/Lovely stuff/);

PR #1002, opened 2026-06-04
Feature/merge diff improvements

jamesrwelch:


edwardcrichton (approved):
Much tidier, both on the interface and in the code. More robust undefined / null checks

PR #1004, opened 2026-06-08
Fixes for the 'search for a catalogue item' dialog box 

jamesrwelch:
- Fix the search by adding a prefix search option (requires update to mdm-resources), and making the dialog box bigger.  
- Use a different endpoint for searching DataElements


edwardcrichton (approved):
🔧 Prefixed fixed

PR #1005, opened 2026-06-08
Fixes for the 'create data element' wizard where creating a data type…

jamesrwelch:
… with enumeration values was not working
Relies on a fix to mauro-micronaut


edwardcrichton (approved):
That looks good. Some important logical changes, *ngIf rather than hidden, if (data.body.items) {...} is a good spot

PR #1006, opened 2026-06-12
Feature/location links

jamesrwelch:
Fixes for links in location / breadcrumbs, and improve the styling


edwardcrichton (approved):
😎 Much better handling of paths and breadcrumbs - parsing and construction. Some better guarding of version tags. 'Branchable' logic. Properly finding the versioned item for hrefs. Much CSS tidying.
Good work!

